PR #4, opened 2018-08-24
Small comma and batch_size placement

nstrayer:
When I was working my way through this (awesome) post I noticed two things. First, when I went to copy and paste the prereq packages I got a confusing error caused by a missing comma. 

Second in the code that gets run the variable `batch_size` is called before it is defined. I added it here where it is needed but I'm not sure it's the best way to do it as it's a tiny bit out of the blue. Let me know if you have another preference (e.g. functionalizing the dataset generation to take `batch_size` as an argument?).

Small note: I simply modified the html file by hand because of how long it takes to run the code. Hopefully that's not a no-no.

### DIFF
--- a/_posts/2018-07-30-attention-layer/attention-layer.Rmd
+++ b/_posts/2018-07-30-attention-layer/attention-layer.Rmd
@@ -40,7 +40,7 @@ The code in this post depends on the development versions of several of the Tens
 
 ```r
 devtools::install_github(c(
-  "rstudio/reticulate"
+  "rstudio/reticulate",
   "rstudio/tensorflow",
   "rstudio/keras",
   "rstudio/tfdatasets"
@@ -228,6 +228,8 @@ This section does not contain much code, but it shows an important technique: th
 Remember the olden times when we used to pass in hand-crafted generators to Keras models? With [tfdatasets](https://tensorflow.rstudio.com/tools/tfdatasets/articles/introduction.html), we can scalably feed data directly to the Keras `fit` function, having various preparatory actions being performed directly in native code. In our case, we will not be using `fit`, instead iterate directly over the tensors contained in the dataset. 
 
 ```{r}
+batch_size <- 32
+
 train_dataset <- 
   tensor_slices_dataset(keras_array(list(x_train, y_train)))  %>%
   dataset_shuffle(buffer_size = buffer_size) %>%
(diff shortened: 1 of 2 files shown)


skeydan:
Hi Nick,

thanks for so much attention reading the attention post :-) !!

Right about the comma, sorry about the confusion, that happened because later I put reticulate first...

About placement of ```batch_size```, I was not expecting people to copy-paste this chunk-by-chunk and therefore I linked to the complete example (https://github.com/rstudio/keras/blob/master/vignettes/examples/nmt_attention.R) where everything is in execution order.
(For example, there the ```evaluate``` and ```plot_attention``` functions are defined before the training loop.)
This (I was thinking) gave me the freedom to introduce things in a "logical order" (although you might say in case of ```batch_size```, logical is where we define the dataset, whereas I was putting it with the NN hyperparams).

But I should have written this right in the beginning - that will be the case in all upcoming posts that have complete, technical examples like this.

BTW it's no problem (and probably better)  to modify the .Rmd and then knit, this will be super fast as ```eval=FALSE```.

jjallaire:
Okay, let's add the comma as is done in this PR and then let's also add a note indicating that readers that want to execute the code should do so from the linked script rather than by copy/paste from the post.

skeydan:
created https://github.com/rstudio/tensorflow-blog/pull/5
thanks again Nick!
@jjallaire the new branches have a similar note before actual code snippets already...

nstrayer:
Shoot, I should have read the article closer! I also noticed the `eval = FALSE` setting after this PR. I was too excited about the code to pay enough attention apparently. 

As a note: the code does execute properly in order with the change of just that batch size variable and then adding print calls to the batch/epoch score messages. I agree though that it takes long enough that the notebook format is probably not the best way to run the code.

nstrayer:
@skeydan I saw you already created a PR w/ the proper fix of the commas. Let me know if I can do anything else! Sorry to have prematurely submitted the PR without realizing I could have knitted easily!

skeydan:
No problem, and thanks again! 

>> Shoot, I should have read the article closer! I 

Not at all, my fault, I should have alerted readers to that right in the beginning! It'll be like that in the upcoming ones on eager execution... (4 to come, at least!)
I'm happy to hear you're excited about it - to me, it really makes some things easier than before!

PR #12, opened 2018-09-17
Eager pix2pix

skeydan:
could you please review? I've been thinking we might publish this on Thursday, and then next week there'll be something non-eager for a change :-)


jjallaire:
Made some minor edits, looks great overall!

The one enhancement I would make it to describe the learning task in a bit more detail at the outset (I was a bit unclear on what the model was actually attempting to do until I had seen a handful of the examples presented later on)

skeydan:
great, thank you!! Let me see what I can do... this is really better seen than described, and definitely I don't want to use the "easy way" out - just giving the formulae - , I want to keep things on a conceptual level in these category of posts

skeydan:
Tried to make it more clear... downside is I'm showing a pic from the paper before talking about the paper, but I'm referring to it & and the way I'm doing it now, the main "order of argumentation" still works & I'm not mixing it all up
Do you think it's ok?

jjallaire:
I think it works great!

skeydan:
wonderful!

PR #13, opened 2018-09-25
ready for review

skeydan:
there it is...


jjallaire:
We might want to address how to save and restore this model (I think we can do so by doing a save/restore of the weights but it would be good to confirm this)

PR #14, opened 2018-10-01
Eager wrapup

skeydan:
this would be ready for review ... thank you!


jjallaire:
I think you should start the post with very big picture motivation (i.e. enumerate the applications demonstrated in the blog posts, not yet linking to the posts mind you, that will come a nice bonus later).

After that, you should define it briefly (as you already do), then note that Eager Execution is available in the R API, linking to the Eager Guide.

Then, get into the explanation (this is largely what the post already has). One change I would make is more prominent linking to the other eager blog posts from within this posts (they are much less prominently linked asides in the current draft).

In your wrap-up section I would once again link to the Eager Guide and the case study blog posts.

PR #16, opened 2018-10-20
ready for review

skeydan:
this one would be ready :-) if you don't have time on Monday, I'll change the date no problem :-)


jjallaire:
I think we need the full source of the example code merged into the keras repo (both of those links 404'd)

PR #22, opened 2018-12-17
Od2

skeydan:


skeydan:
thanks so much for reading! I'll fix and merge :-)

PR #32, opened 2019-03-15
ready for review

skeydan:
Hi JJ, could you please review?
I wonder if there is a place to attribute the thumbnail image, taken from 
https://commons.wikimedia.org/wiki/File:A-young-girl-reading_m_ancher.jpg
(on that page they say it's not legally required, but then, we have this

> The figures that have been reused from other sources don't fall under this license and can be recognized by a note in their caption: "Figure from ...".

in our generic footer...

Also I 'm wondering if there is a way to resize the thumbnail (if I download a low-res version, it gets scaled up).

Thanks!!


jjallaire:
Post looks great! I would actually crop the thumbnail so it's fits better alongside the summary text on the home page. Don't worry about attribution for the thumbnail since it's not actually in the post.

skeydan:
great idea, thanks! The image is still quite big but it has the right aspect ratio now.

PR #41, opened 2019-06-04
ready for review

skeydan:


_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,779 @@
...
+We re-use the dataset from the Google TensorFlow Probability team's [blog post on the same subject]( https://medium.com/tensorflow/regression-with-probabilistic-layers-in-tensorflow-probability-e46ff5d37baf) ^[see also the corresponding [notebook](https://colab.research.google.com/github/tensorflow/probability/blob/master/tensorflow_probability/examples/jupyter_notebooks/Probabilistic_Layers_Regression.ipynb#scrollTo=5zCEYpzu7bDX)], with one exception: We extend the range of the independent variable a bit on the negative side, to better demonstrate the different methods' behaviors.
+
+Here is the data-generating process. We also get library loading out of the way. Like in the preceding posts on `tfprobability`, please use the development versions of `tensorflow` and `tfprobability` as well as `keras`. Call `install_tensorflow(version = "nightly")` to obtain a current nightly build of TensorFlow and TensorFlow Probability:
+
+
+```{r}
+# make sure we use the development versions of tensorflow, tfprobability and keras
+devtools::install_github("rstudio/tensorflow")
+devtools::install_github("rstudio/tfprobability")
+devtools::install_github("rstudio/keras")
```
javierluraschi: Why do we need `keras` and `tensorflow` devel? I guess to use nightly?

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,779 @@
...
+
+# and that we use a nightly build of TensorFlow and TensorFlow Probability
+tensorflow::install_tensorflow(version = "nightly")
+
+library(tensorflow)
+tf$compat$v1$enable_v2_behavior()
```
javierluraschi: Do you really need to `enable_v2_behavior()`? I guess `tfprobabilty` requires it? You might want to consider using TF 2.0 in subsequent posts, assuming the fixes that you need are being ported to TF 2.0, is that the case?

kevinykuo:
LGTM.

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
skeydan: there've been quite a few changes since the CRAN release, one of which involves load order of packages (keras/tensorflow), i'm pretty sure it's safest to use nightly for all 3 here...

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
skeydan: In TFP, they just do this all the time since a few months... I think here too, it's the safest way.
Also, now that TF 2 beta is imminent, seems like it won't be long for TF 2 anyway...

javierluraschi: I wonder if you should consider adding a:

```
.onLoad <- function() {
   # if tf < 2.0
   tf$compat$v1$enable_v2_behavior()
}
```

To `tfprobability`, that way you can get rid of the `tf$compat$v1$enable_v2_behavior()` statement.

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
javierluraschi: If you still have the error around, I would share it to make sure there is not something "broken" in tensorflow/keras, would be nice to understand why we can't use `tfprobability` with older TensorFlow versions... or have `tf.probability` already state they are not backwards compatible and the latest version of TF is required?

skeydan: Sorry I was unclear, I guess it's 2 different things:

1. we need to push some changes (that loading order one especially) to CRAN for seemless integration of `keras` and `tfprobability` (the R packages)

2. TFP (Python) always needs a corresponding build of TF (so, nightly needs nightly). And the functionality I use is not the - pretty old! - TFP release 0.6.

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
skeydan: Again I was unclear, `tfprobability` per se doesn't need it - the honest way to express this would probably be "it's the idiomatic way to use TF master nowadays". It 's also a way to make sure the code will (hopefully!) work with the final TF 2.

javierluraschi: I would then consider moving clean up the code a bit as follows:

```
library(tensorflow)
library(tfprobability)
library(keras)

library(dplyr)
library(tidyr)
library(ggplot2)

# make sure this code is compatible with TensorFlow 2.0
tf$compat$v1$enable_v2_behavior()
```

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
javierluraschi: Gotcha! So the post itself requires nightly TFP. So I would change this:

> Like in the preceding posts on tfprobability, please use the development versions of tensorflow and tfprobability as well as keras.

to

> Like in the preceding posts on tfprobability, this post also makes use of new tfprobability features, so please use the development versions of tensorflow and tfprobability as well as keras.

kevinykuo:
Actually, couple things to consider

1. Since the main goal is to talk about uncertainty, we could just talk about linear models (i.e. get rid of the hidden layer.)
2. Start with an example of the homoscedastic approach, i.e.

```
model <- keras_model_sequential() %>%
  # layer_dense(units = 8, activation = "relu") %>%
  layer_dense(units = 1)
```

so we contrast getting point estimates (or more correctly, distributions with constant variance) vs. distributions.

skeydan:
Thanks @kevinykuo! Yes I did consider these, but in the end decided otherwise (so many decisions to make also in writing this post!):

For (1), yes I see the point, but I can imagine the aleatoric way ending up as the most used (in the beginning, at least), as it's straightforward, transparent and simple (in spite of its simplicity). So I wanted to have an example that shows you can just use it with any NN (not just unrealistic, linear ones). In that same spirit, I tend to talk less of uncertainty than of "capturing the spread of the data" when referring to that method.

Reg. (2), yes I see that option too but I don't think I need it for the story (which is built more on picking up on that previous post on the same topic).

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
skeydan: done!

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
skeydan: done!

kevinykuo:
How come for the variational example we don't have the hidden layer?

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
+---
+title: "Adding uncertainty estimates to Keras models with tfprobability"
+description: > 
+   As of today, there is no mainstream road to obtain uncertainty estimates from neural networks. All that can be said this that normally, approaches tend to be Bayesian in spirit, involving some way of putting a prior over model weights. This holds true as well for the method presented in this post: We show how to use tfprobability, the R interface to TensorFlow Probability, to add uncertainty estimates to a Keras model in an elegant and conceptually plausible way.
```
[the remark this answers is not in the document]
batpigandme: I think you might cut _This holds true as well for the method presented in this post:_ from the above, since it's implied that you're describing what's contained in the post itself.

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
...
+preview: images/uci_both.png
+---
+
+
+```{r setup, include=FALSE}
+knitr::opts_chunk$set(echo = TRUE, eval = FALSE)
+```
+
+
+About six months ago, we showed [how to create a custom wrapper](https://blogs.rstudio.com/tensorflow/posts/2018-11-12-uncertainty_estimates_dropout/) to obtain uncertainty estimates from a Keras network. Today we present a less laborious, as well faster-running, way using [tfprobability](https://rstudio.github.io/tfprobability/), the R wrapper to TensorFlow Probability. Like most posts on this blog, this one won't be short, so let's quickly state what you can expect in return of reading time.
```
batpigandme:
```suggestion
About six months ago, we showed [how to create a custom wrapper](https://blogs.rstudio.com/tensorflow/posts/2018-11-12-uncertainty_estimates_dropout/) to obtain uncertainty estimates from a Keras network. Today we present a less laborious, as well faster-running way using [tfprobability](https://rstudio.github.io/tfprobability/), the R wrapper to TensorFlow Probability. Like most posts on this blog, this one won't be short, so let's quickly state what you can expect in return of reading time.
```

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
...
+
+## What to expect from this post
+
+Starting from what _not_ to expect: There won't be a recipe that tells you how _exactly_ to set all parameters involved in order to report the "right" uncertainty measures. But then, what _are_ the "right" uncertainty measures? Unless you happen to work with a method that has no (hyper-)parameters to tweak, there will always be uncertainty about how to report uncertainty.
```
batpigandme: _there will always be uncertainty about how to report uncertainty._
Do you mean uncertainty in the first mention, here? Is it something closer to ambiguity/no one right way to do it?

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
...
+
+What you _can_ expect though is an introduction to obtaining uncertainty estimates for Keras networks, as well as an empirical report of how tweaking (hyper-)parameters may affect the results. As in the aforementioned post, we perform our tests on a simulated as well as on a real dataset, the [Combined Cycle Power Plant Data Set](http://archive.ics.uci.edu/ml/datasets/Combined+Cycle+Power+Plant). At the end, in place of strict rules you should have acquired some intuition that will transfer to other real-world datasets.
```
batpigandme:
```suggestion
What you _can_ expect, though, is an introduction to obtaining uncertainty estimates for Keras networks, as well as an empirical report of how tweaking (hyper-)parameters may affect the results. As in the aforementioned post, we perform our tests on both a simulated and a real dataset, the [Combined Cycle Power Plant Data Set](http://archive.ics.uci.edu/ml/datasets/Combined+Cycle+Power+Plant). At the end, in place of strict rules, you should have acquired some intuition that will transfer to other real-world datasets.
```

batpigandme: Restructured the second sentence to use "and" instead of "as well as" (since you used it just above). Also added a comma after "in place of strict rules". 
😳 Sorry, I don't know why the GitHub UI doesn't show specific diffs in some cases.

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
...
+
+Did you notice our talking about _Keras networks_ above? Indeed this post has an additional goal: So far, we haven't really discussed yet how `tfprobability` goes together with `keras`. Now we finally do (in short: they work together seemlessly).
+
+Finally, the notions of _aleatoric_ and _epistemic_ uncertainty, which may have stayed a bit abstract in the prior post, should get much more concrete here.
```
batpigandme: 🖤

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
...
+
+## Aleatoric vs. epistemic uncertainty
+
+Reminiscent somehow of the classic decomposition of generalization error into bias and variance, splitting uncertainty into its epistemic and aleatoric constituents separates an irreducible from a reducible part.
```
batpigandme: If there's a piece somewhere that you like on decomposition of generalization error, it might be nice to have a recommended reading/link here. If not, it's not a big deal!

batpigandme: Also, maybe add hover-able definitions for epistemic and aleatoric? I've run into quite a few folks unfamiliar with the latter term (I knew it from aleatoric music — good old John Cage's 4'33 of silence, for one). I don't think you need anything big, but it's nice to throw the reader a bone, and keep them from having to look it up!

batpigandme: You, in effect, define them below, but I also think they're just cool words!

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
...
+
+The reducible part relates to imperfection in the model: In theory, if our model were perfect, epistemic uncertainty would vanish. Put differently, if the training data were unlimited -- or if they comprised the whole population -- we could just add capacity to the model until we've obtained a perfect fit.
+
+In contrast, normally there is variation in our measurements. There may be one true process that determines my resting heart rate; nonetheless, actual measurements will vary over time. There is nothing to be done about this: This is the aleatoric part that just remains, to be factored into our expectations.
+
+Now reading this, you might be thinking: "Wouldn't a model that actually _were_ perfect capture those pseudo-random fluctuations?". We'll leave that phisosophical question be;  instead, we'll try to illustrate the usefulness of this distinction by example, in a practical way. In a nutshell, viewing a model's _aleatoric_ uncertainty output should caution us to factor in appropriate deviations when making our predictions, while inspecting _epistemic_ uncertainty should help us re-think the appropriateness of the chosen model.
```
batpigandme:
```suggestion
Now reading this, you might be thinking: "Wouldn't a model that actually _was_ perfect capture those pseudo-random fluctuations?". We'll leave that philosophical question be;  instead, we'll try to illustrate the usefulness of this distinction by example, in a practical way. In a nutshell, viewing a model's _aleatoric_ uncertainty output should caution us to factor in appropriate deviations when making our predictions, while inspecting _epistemic_ uncertainty should help us rethink the appropriateness of the chosen model.
```

batpigandme: ^^ changes were to was (since model is singular), corrects spelling of philosophical and rethink.

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
...
+ggplot(data.frame(x = x, y = y), aes(x, y)) + geom_point()
+```
+
+
+```{r, eval=TRUE, echo=FALSE, layout="l-body-outset", fig.cap = "Simulated data"}
+knitr::include_graphics("images/simdata.png")
+```
+
+
+The task here is single-predictor regression, which in principle we can achieve use Keras `dense` layers.
```
batpigandme:
```suggestion
The task here is single-predictor regression, which in principle we can achieve using Keras `dense` layers.
```

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
...
+Let's see how to enhance this by indicating uncertainty, starting from the aleatoric type.
```
batpigandme:
```suggestion
Let's see how to enhance this by indicating uncertainty, starting with the aleatoric type.
```

batpigandme: ^^ I guess this could be either, but my Sprachgefühl pings with. Feel free to ignore!

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
...
+
+
+### Aleatoric uncertainty
+
+Aleatoric uncertainty, by definition, is not a statement about the model. So why not have the model learn the uncertainty inherent in the data?
```
batpigandme:
```suggestion
Aleatoric uncertainty, by definition, is not a statement about the model. So, why not have the model learn the uncertainty inherent in the data?
```

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
...
+
+This is exactly how aleatoric uncertainty is operationalized in this approach. Instead of a single output per input -- the predicted mean of the regression -- here we have two outputs: one for the mean, and one for the standard deviation.
+
+How will we use these? Until shortly, we would have had to roll our own logic. Now with `tfprobability`, we make the network output not tensors, but distributions -- put differently, we make the last layer a [distribution layer](https://rstudio.github.io/tfprobability/reference/index.html#section-distribution-layers).
```
batpigandme:
```suggestion
How will we use these? Until shortly, we would have had to roll our own logic. Now, with `tfprobability`, we make the network output not tensors, but distributions -- put differently, we make the last layer a [distribution layer](https://rstudio.github.io/tfprobability/reference/index.html#section-distribution-layers).
```

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
...
+We can now compile and fit the model.
+
+```{r}
+learning_rate <- 0.01
+model %>% compile(optimizer = optimizer_adam(lr = learning_rate), loss = negloglik)
+
+model %>% fit(x, y, epochs = 1000)
+```
+
+We now call the model on the test data to obtain the predictions. The predictions now actually are _distributions_, and we have 150 of them, one for each datapoint:
```
batpigandme:
```suggestion
We now call the model on the test data to obtain the predictions. The predictions actually are _distributions_, and we have 150 of them, one for each datapoint:
```

batpigandme: Removed the second "now," but the first could go instead, if you prefer.

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
...
+
+```
+
+This time, the model does not capture the "form" of the data that well, as we've disallowed any nonlinearities.
+
+```{r, eval=TRUE, echo=FALSE, layout="l-body-outset", fig.cap = "Aleatoric uncertainty on simulated data, using linear activation in the first dense layer."}
+knitr::include_graphics("images/g_aleatoric_linear_8.png")
+```
+
+Using linear activations only, we also need to do more experimenting with the `scale = ...` line to get the result look "right". With `relu`, on the other hand, results are pretty robust to changes in how `scale` is computed. Which activation do we choose? If our goal is to adequately model variation in the data, we can just choose `relu` -- and leave assessing _uncertainty in the model_ to a different technique (the _epistemic_ uncertainty that is up next).
```
batpigandme:
```suggestion
Using linear activations only, we also need to do more experimenting with the `scale = ...` line to get the result to look "right". With `relu`, on the other hand, results are pretty robust to changes in how `scale` is computed. Which activation do we choose? If our goal is to adequately model variation in the data, we can just use `relu` -- and leave assessing _uncertainty in the model_ to a different technique (the _epistemic_ uncertainty that is up next).
```

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
...
+
+Overall, it seems like aleatoric uncertainty is the straightforward part. We want the network to learn the variation inherent in the data, which it does. What do we gain? Instead of obtaining just point estimates, which in this example might turn out pretty bad in the two fan-like areas of the data on the left and right sides, we learn about the spread as well. We'll thus be appropriately cautious depending on what input range we're making predictions for.
```
batpigandme:
```suggestion
Overall, it seems like aleatoric uncertainty is the straightforward part. We want the network to learn the variation inherent in the data, which it does. What do we gain? Instead of obtaining just point estimates, which in this example might turn out pretty bad in the two fan-like areas of the data on the left and right sides, we learn about the spread as well. We'll thus be appropriately cautious, depending on what input range we're making predictions for.
```

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
...
+
+### Epistemic uncertainty
+
+Now our focus is on the model. Given a speficic model (e.g., one from the linear family), what kind of data does it say conforms to its expectations?
+
+To answer this question, we make use of a [variational-dense layer](https://rstudio.github.io/tfprobability/reference/layer_dense_variational.html).
+This is again a Keras layer provided by `tfprobability`. Internally, it works by minimizing the _evidence lower bound_ (ELBO),  thus striving to find an approximative posterior that does two things:
```
batpigandme:
```suggestion
This is, again, a Keras layer provided by `tfprobability`. Internally, it works by minimizing the _evidence lower bound_ (ELBO),  thus striving to find an approximative posterior that does two things:
```

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
...
+      layer_distribution_lambda(function(t) {
+        tfd_independent(tfd_normal(loc = t, scale = 1),
+                        reinterpreted_batch_ndims = 1)
+      })
+  }
+
+```
+
+
+This prior is itself a Keras model, containing a [ layer that wraps a variable](https://rstudio.github.io/tfprobability/reference/layer_variable.html) and a `layer_distribution_lambda`, that type of distribution-yielding layer we've just encountered above. The variable layer could be fixed (non-trainable) or non-trainable, corresponding to a genuine prior or a prior learnt from the data in an _empirical Bayes_-like way. The distribution layer outputs a normal distribution since we're in a regression setting.
```
batpigandme: > fixed (non-trainable) or non-trainable

This is a little confusing to me. Are the two options: (1) fixed non-trainable, (2) non-trainable?

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
...
+
+The posterior too is a Keras model -- definitely trainable this time. It too outputs a normal distribution:
```
batpigandme:
```suggestion
The posterior is also a Keras model -- definitely trainable this time. It too outputs a normal distribution:
```

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
...
+            loc = t[1:n],
+            scale = 1e-5 + tf$nn$softplus(c + t[(n + 1):(2 * n)])
+            ), reinterpreted_batch_ndims = 1)
+        }
+      )
+    ))
+  }
+```
+
+Now that we've defined both, we can set up the model's layers. The first one, a variational-dense layer, has a single unit. The ensuing distribution layer then takes that unit's output and uses it for the mean of a normal distribution -- while the scale of that Normal is fixed at 1:
```
batpigandme: Don't know if the Normal with a capital N at the end is a convention or a typo!

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
...
+  theme(legend.position = "none")
+```
+
+```{r, eval=TRUE, echo=FALSE, layout="l-body-outset", fig.cap = "Epistemic uncertainty on simulated data, using linear activation in the variational-dense layer."}
+knitr::include_graphics("images/g_epistemic_linear_kl150.png")
+```
+
+What we see here are essentially _different models_, consistent with the assumptions built into the architecture. What we're not accounting for is the spread in the data. Can we do both? We can; but first let's comment on a few choices that were made and see how they affect the results. 
+
+To prevent this post from growing to infinite size, we've refrained from performing a systematic experiment; please take what follows not as generalizable statements, but as _pointers to things you will want to keep in mind_ in your own ventures. Especially, each (hyper-)parameter is not an island; they could interact in unforeseen ways.
```
batpigandme: I wonder if it would be worth having a subheading for this section on questions/things to pay attention to.

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
...
+
+
+```{r, eval=TRUE, echo=FALSE, layout="l-body-outset", fig.cap = "Epistemic uncertainty on simulated data. Left: kl_weight = 1. Right: prior non-trainable."}
+knitr::include_graphics("images/g_epistemic_100epochs.png")
+```
+
+
+Now let's come back to the question: We've modeled spread in the data, we've peeked into the heart of the model, -- can we do both at the same time? 
+
+We can if we combine both approaches. We add an additional unit to the variational-dense layer and use this to learn the variance: once for each "sub-model" contained in the model.
```
batpigandme:
```suggestion
We can, if we combine both approaches. We add an additional unit to the variational-dense layer and use this to learn the variance: once for each "sub-model" contained in the model.
```

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
...
+```
+
+Evidently, this is another parameter we should be prepared to experiment with.
+
+Now that we've introduced all three types of presenting uncertainty -- aleatoric only, epistemic only, or both -- let's see them on the aforementioned [Combined Cycle Power Plant Data Set](http://archive.ics.uci.edu/ml/datasets/Combined+Cycle+Power+Plant). Please see [our previous post on uncertainty](https://blogs.rstudio.com/tensorflow/posts/2018-11-12-uncertainty_estimates_dropout/) for a quick characterization, as well as visualization, of the dataset.
+
+## Combined Cycle Power Plant Data Set
+
+
+To keep this post at a digestible length, we'll refrain from trying as many alternatives as with the simulated data and mainly stay with what worked well there. This should also give us an idea of how well these "defaults" generalize. We separately inspect two scenarios: The single-predictor setup (using each of the four available predictors alone) and the complete one (using all four predictors at once).
```
batpigandme:
```suggestion
To keep this post at a digestible length, we'll refrain from trying as many alternatives as with the simulated data and mainly stay with what worked well there. This should also give us an idea of how well these "defaults" generalize. We separately inspect two scenarios: The single-predictor setup (using each of the four available predictors alone), and the complete one (using all four predictors at once). 
```

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
...
+  theme(legend.position = "none")
+```
+
+And this is the result. 
+
+```{r, eval=TRUE, echo=FALSE, layout="l-body-outset", fig.cap = "Epistemic uncertainty on the Combined Cycle Power Plant Data Set; single predictors."}
+knitr::include_graphics("images/uci_epistemic.png")
+```
+
+As with the simulated data, the linear models seems to "do the right thing". And here too, we think we will want to augment this with the spread in the data: Thus, on to way 3.
```
batpigandme: > Thus, on to way 3.

I would write out the word three, here.

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
```diff
@@ -0,0 +1,782 @@
...
+
+```{r, eval=TRUE, echo=FALSE, layout="l-body-outset", fig.cap = "Combined uncertainty on the Combined Cycle Power Plant Data Set; single predictors."}
+knitr::include_graphics("images/uci_both.png")
+```
+
+This looks useful! Let's wrap up with our final test case: Using all four predictors together.
+
+### All predictors
+
+The training code used in this scenario looks just like before, apart from our feeding all predictors to the model. For plotting, we resort to displaying the first principal component on the x-axis -- this makes the plots look noisier than before. We also display fewer lines for the epistemic and epistemic plus aleatoric cases (20 instead of 100). Here are the results:
```
batpigandme:
```suggestion
The training code used in this scenario looks just like before, apart from our feeding all predictors to the model. For plotting, we resort to displaying the first principal component on the x-axis -- this makes the plots look noisier than before. We also display fewer lines for the epistemic and epistemic-plus-aleatoric cases (20 instead of 100). Here are the results:
```

batpigandme: ^^ added hyphens since it's being used as a compound modifier for "cases"

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
skeydan: I think I wanted it to be some kind of pun, but I think this might not be working correctly in English... the first "uncertainty" should have a different connotation, like "not feeling sure about"..."incertitude", perhaps? Or just: "questions"?

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
skeydan: I'd rather not over-stress these, for pragmatic reasons...

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
skeydan: oh, right, trying to do that consistently otherwise...

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
skeydan: oh, I was wondering how to write that, that's nice!

_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd
skeydan: went with questions now, think I'm ready to merge then...

skeydan:
@kevinykuo 

> How come for the variational example we don't have the hidden layer?

commenting on that here: https://github.com/rstudio/tensorflow-blog/blob/6dee441070781df43048450713f62d311753d214/_posts/2019-06-05-uncertainty-estimates-tfprobability/uncertainty_tfprobability.Rmd#L371

skeydan:
Thanks a lot all, will merge...

PR #45, opened 2019-07-29
ready for review

skeydan:


_posts/2019-07-30-censored-data/censored_data_tfp.Rmd
```diff
@@ -0,0 +1,724 @@
...
+---
+
+```{r setup, include=FALSE}
+knitr::opts_chunk$set(echo = TRUE, eval = FALSE)
+```
+
+Nothing's ever perfect, and data isn't either. One type of "imperfection" is _missing data_, where some features are unobserved for some subjects. (A topic for another post.) Another is _censored data_, where an event whose characteristics we want to measure does not occur in the observation interval. The example in Richard McElreath's _Statistical Rethinking_ is time to adoption of cats in an animal shelter. If we fix an interval and observe wait times for those cats that actually _did_ get adopted, our estimate will end up too optimistic: We don't take into account those cats who weren't adopted during this interval and thus, would have contributed wait times of length longer than the complete interval.
+
+In this post, we use a slightly less emotional example which nonetheless may be of interest, especially to R package developers: time to completion of `R CMD check`, collected from CRAN and provided by the `parsnip` package as `check_times`. Here, the censored portion are those checks that errored out for whatever reason, i.e., for whom the check did not complete.
```
batpigandme:
```suggestion
In this post, we use a slightly less emotional example which nonetheless may be of interest, especially to R package developers: time to completion of `R CMD check`, collected from CRAN and provided by the `parsnip` package as `check_times`. Here, the censored portion are those checks that errored out for whatever reason, i.e., for which the check did not complete.
```

_posts/2019-07-30-censored-data/censored_data_tfp.Rmd
```diff
@@ -0,0 +1,724 @@
...
+
+Why do we care about the censored portion, the entities for whom the event of interest did not occur in the period of observation? In the cat adoption scenario, this is pretty obvious: We want to be able to get a realistic estimate for any unknown cat, not just those cats that will turn out to be "lucky". How about `check_times`? Well, if your submission is one of those that errored out, you still care about how long you wait, so even though their percentage is low (< 1%) we don't want to simply exclude them. Also, there is the possibility that the failing ones would have taken longer, had they run to completion, dut to some intrinsic difference between both groups. Conversely, if failures were random, the longer-running checks would have a greater chance to get hit by an error. So here too, exluding the censored data may result in bias.
```
batpigandme: I would open this with the question: "Why do we care about the censored portion?" Then describe the cat scenario (as you do now), _then_ go into the check times/entities for which event of interest didn't occur.

_posts/2019-07-30-censored-data/censored_data_tfp.Rmd
```diff
@@ -0,0 +1,724 @@
...
+library(parsnip)
+library(tidyverse)
+library(zeallot)
+library(gridExtra)
+library(HDInterval)
+
+tf$compat$v2$enable_v2_behavior()
+```
+
+Besides the check durations we want to model, `check_times` reports various features of the package in question, such as number of imported packages, number of dependencies, size of code and documentation files, etc. The `status` variable indicates whether the check completed or errored out.
```
batpigandme:
```suggestion
Besides the check durations we want to model, `check_times` reports' various features of the package in question, such as number of imported packages, number of dependencies, size of code and documentation files, etc. The `status` variable indicates whether the check completed or errored out.
```

_posts/2019-07-30-censored-data/censored_data_tfp.Rmd
```diff
@@ -0,0 +1,724 @@
...
+$ src_size       <dbl> 0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000,…
+$ data_count     <int> 2, 0, 0, 3, 3, 1, 10, 0, 4, 2, 2, 146, 0, 0, 0, 0, 0, 10, 0…
+$ data_size      <dbl> 0.025292, 0.000000, 0.000000, 4.885864, 4.595504, 0.006500,…
+$ testthat_count <int> 0, 8, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 3, 3, 0, 0,…
+$ testthat_size  <dbl> 0.000000, 0.002496, 0.000000, 0.000000, 0.000000, 0.000000,…
+$ check_time     <dbl> 49, 101, 292, 21, 103, 46, 78, 91, 47, 196, 200, 169, 45, 2…
+$ status         <dbl> 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,…
+```
+
+Of these 13626 observations, just 103 are censored:
```
batpigandme:
```suggestion
Of these 13,626 observations, just 103 are censored:
```

_posts/2019-07-30-censored-data/censored_data_tfp.Rmd
```diff
@@ -0,0 +1,724 @@
...
+```{r}
+table(df$status)
+```
+
+```
+0     1 
+103 13523 
+```
+
+For better readability of this text, we want to work with an interesting subset of the columns. We use `lm` to help us guide our selection:
```
batpigandme:
```suggestion
For better readability, we'll work with an interesting subset of the columns. We use `lm` to help us guide our selection:
```

_posts/2019-07-30-censored-data/censored_data_tfp.Rmd
```diff
@@ -0,0 +1,724 @@
...
+---
+Signif. codes:  0 ‘***’ 0.001 ‘**’ 0.01 ‘*’ 0.05 ‘.’ 0.1 ‘ ’ 1
+
+Residual standard error: 76.6 on 13602 degrees of freedom
+Multiple R-squared:  0.4282,	Adjusted R-squared:  0.4273 
+F-statistic: 442.9 on 23 and 13602 DF,  p-value: < 2.2e-16
+
+```
+
+It seems that if we choose `imports`, `depends`, `r_size`, `doc_size`, `ns_import` and `ns_export` we end up with a mix of (comparatively) powerful predictors from different semantic spaces and of different scales.
```
batpigandme:
```suggestion
It seems that if we choose `imports`, `depends`, `r_size`, `doc_size`, `ns_import` and `ns_export`, we end up with a mix of (comparatively) powerful predictors from different semantic spaces and of different scales.
```

_posts/2019-07-30-censored-data/censored_data_tfp.Rmd
```diff
@@ -0,0 +1,724 @@
...
+  as.matrix()
+
+# check times for successful checks 
+check_time_nc <- df %>%
+  filter(status == 1) %>%
+  select(check_time) %>%
+  as.matrix()
+```
+
+Now we can zoom in on the variables of interest, setting up one dataframe for the censored data and one for the uncensored data each. We add a column of `1`s for use as an intercept.
```
batpigandme:
```suggestion
Now we can zoom in on the variables of interest, setting up one data frame for the censored data and one for the uncensored data. We add a column of `1`s for use as an intercept.
```

_posts/2019-07-30-censored-data/censored_data_tfp.Rmd
```diff
@@ -0,0 +1,724 @@
...
+And finally, we see it for `doc_size` too, where `doc_size` is the size of `.Rmd` and `.Rnw` files:
+
+::: l-body
+| percentiles: _doc_size_     |   10%   |   30%   |   50%   |   70%   |   90%  |    100%   |
+|-----------------------------|---------|---------|---------|---------|--------|-----------|
+| completed                   |   0.000 |   0.000 |   0.000 |   0.000 |  0.023 |     0.988 |
+| not completed               |   0.000 |   0.000 |   0.000 |   0.011 |  0.042 |     0.114 |
+:::
+
+Given our task at hand -- model check durations taking into account uncensored as well as censored data -- we won't dwell on differences between both groups any longer; nonetheless we thought it interesting to relate these numbers.
```
batpigandme:
```suggestion
Given the task at hand -- model check durations taking into account uncensored as well as censored data -- we won't dwell on differences between both groups any longer; nonetheless we thought it interesting to relate these numbers.
```

_posts/2019-07-30-censored-data/censored_data_tfp.Rmd
```diff
@@ -0,0 +1,724 @@
...
+
+So now, back to work. We need to create a model.
+
+## The model
+
+As explained in the introduction, for completed checks duration is modeled using an exponential PDF. This is as straightforward as adding [tfd_exponential()](https://rstudio.github.io/tfprobability/reference/tfd_exponential.html) to the model function, [tfd_joint_distribution_sequential()](https://rstudio.github.io/tfprobability/reference/tfd_joint_distribution_sequential.html). ^[For a first introduction to MCMC sampling with `tfprobability`, see [Tadpoles on TensorFlow: Hierarchical partial pooling with tfprobability](https://blogs.rstudio.com/tensorflow/posts/2019-05-06-tadpoles-on-tensorflow/)]For the censored portion, we need the exponential CCDF. This one is not, as of today, easily added to the model. What we can do though is calculate its value ourselves and add it to the "main" model likelihood. We'll see this below when discussing sampling; for now it means the model definition ends up straightforward as it only covers the non-censored data. It is made of just the said exponential PDF and priors for the regression parameters.
```
batpigandme: I think this is the first time you've used PDF here, so you might want to spell it out as well.

batpigandme:
```suggestion
As explained in the introduction, for completed checks duration is modeled using an exponential PDF. This is as straightforward as adding [tfd_exponential()](https://rstudio.github.io/tfprobability/reference/tfd_exponential.html) to the model function, [tfd_joint_distribution_sequential()](https://rstudio.github.io/tfprobability/reference/tfd_joint_distribution_sequential.html). ^[For a first introduction to MCMC sampling with `tfprobability`, see [Tadpoles on TensorFlow: Hierarchical partial pooling with tfprobability](https://blogs.rstudio.com/tensorflow/posts/2019-05-06-tadpoles-on-tensorflow/)]For the censored portion, we need the exponential CCDF. This one is not, as of today, easily added to the model. What we can do though is calculate its value ourselves and add it to the "main" model likelihood. We'll see this below when discussing sampling. For now it means the model definition ends up straightforward as it only covers the non-censored data. It is made of just the exponential PDF and priors for the regression parameters.
```

batpigandme: Removed extraneous "the" from last sentence

_posts/2019-07-30-censored-data/censored_data_tfp.Rmd
```diff
@@ -0,0 +1,724 @@
...
+
+As to the latter, we use 0-centered, wide, Gaussian priors for all parameters. As the priors are all the same, instead of listing a bunch of `tfd_normal`s, we can create them all at once as
```
batpigandme:
```suggestion
As for the latter, we use 0-centered, wide, Gaussian priors for all parameters. As the priors are all the same, instead of listing a bunch of `tfd_normal`s, we can create them all at once as
```

_posts/2019-07-30-censored-data/censored_data_tfp.Rmd
```diff
@@ -0,0 +1,724 @@
...
+```{r}
+unconstraining_bijectors <- list(tfb_exp(), tfb_identity())
+```
+
+The next thing to do is define the target we want to optimize.
+
+## Optimization target
+
+Abstractly, the thing to maximize is the log probility of the data -- that is, the measured durations -- under the model.
+Now here the data comes in two parts, and the target does as well. First, we have the non-censored data, for whom
```
batpigandme:
```suggestion
Now here the data comes in two parts, and the target does as well. First, we have the non-censored data, for which
```

_posts/2019-07-30-censored-data/censored_data_tfp.Rmd
```diff
@@ -0,0 +1,724 @@
...
+  e <-  tfd_independent(tfd_exponential(rate = 1 / tf$transpose(tf$matmul(
+    tf$cast(data, betas$dtype), tf$transpose(betas)
+  ))),
+  reinterpreted_batch_ndims = 1)
+  cum_prob <- e %>% tfd_cdf(tf$cast(target, betas$dtype))
+  tf$math$log(1 - cum_prob)
+}
+```
+
+Both parts are combined in a little wrapper function that allows to compare training including and excluding the censored data. We won't do that in this post, but you might be interested to do it with your own data, especially if the ratio of censored and uncensored parts is a little less imbalanced.
```
batpigandme:
```suggestion
Both parts are combined in a little wrapper function that allows us to compare training including and excluding the censored data. We won't do that in this post, but you might be interested to do it with your own data, especially if the ratio of censored and uncensored parts is a little less imbalanced.
```

_posts/2019-07-30-censored-data/censored_data_tfp.Rmd
```diff
@@ -0,0 +1,724 @@
...
+
+We then convert the `samples` tensor to an R array for use in postprocessing.
+
+```{r}
+# 2-item list, where each item has dim (1000, 4)
+samples <- as.array(samples) %>% array_branch(margin = 3)
+```
+
+
+Now, how well did the sampling work? The chains mix well, but for some parameters, autocorrelation is still pretty high, individual tuning of step sizes nonwithstanding.
```
batpigandme:
```suggestion
How well did the sampling work? The chains mix well, but for some parameters, autocorrelation is still pretty high, individual tuning of step sizes notwithstanding.
```

_posts/2019-07-30-censored-data/censored_data_tfp.Rmd
```diff
@@ -0,0 +1,724 @@
...
+```
+
+[1]  9273.062 11328.088
+```
+
+Overall, we conclude that the model works reasonably well.
+
+## Wrapup
+
+We've shown how to model censored data -- resp. a frequent subtype thereof involving durations -- using `tfprobability`. The `check_times` data from `parsnip` were a fun choice, but this modeling technique may be even more useful when censoring is more substantial. Hopefully his post has provided some guidance on how to handle censored data in your own work. Thanks for reading!
```
batpigandme: What does `resp.` stand for here?

_posts/2019-07-30-censored-data/censored_data_tfp.Rmd
skeydan: I think this is a misunderstanding - "reports" is supposed to be read as a verb - had I better choose a different one?

_posts/2019-07-30-censored-data/censored_data_tfp.Rmd
skeydan: respectively - I wasn't sure either if this was okay ... I was looking for an alternative to "or" ... would you have an idea?

_posts/2019-07-30-censored-data/censored_data_tfp.Rmd
batpigandme: (>ლ) No, that makes total sense — I think I wasn't reading in "code mode"!

skeydan: okay thanks!

_posts/2019-07-30-censored-data/censored_data_tfp.Rmd
skeydan: I could change it to "or rather, a frequent subtype..."?
Would that be better?

batpigandme: Yes!

_posts/2019-07-30-censored-data/censored_data_tfp.Rmd
skeydan: Hm, I think that is the order I'm following?
I mean, 

"the entities for whom the event of interest did not occur in the period of observation?"

is really nothing than an "extended paraphrase"?

_posts/2019-07-30-censored-data/censored_data_tfp.Rmd
skeydan: Okay thanks!

skeydan:
Thanks a 1000 times Mara! :-)

_posts/2019-07-30-censored-data/censored_data_tfp.Rmd
batpigandme: Yeah — it's extended enough that it makes it a bit confusing to have the cats in the middle, and then go back to it.

skeydan: Yeah you're right, I'm probably overdoing it rephrasing things times & again ... I  think I'll just remove the paraphrase completely, right?

kevinykuo:
Good stuff! Just a few minor comments
- When I see "censored" in the context of check times the first concept that comes to mind is "the check eventually completes, but it timed out so we don't know how long it actually takes." It seems like this isn't what's being considered, i.e. if it errors right away for an unknown reason we also treat it as censored. Might be worth a couple sentences to clarify this (though I might be the only one who had trouble with this.)
- The quantiles are helpful, but it'd be useful to include a couple histogram plots I think as part of the "EDA".
- Maybe expand on how we're using `lm()` for variable selection so we're not implicitly endorsing variable selection by p-values.
- Should we compare to a `surv_reg()` instead of `lm()`?

topepo:
Small typo:
> Also, there is the possibility that the failing ones would have taken longer, had they run to completion, **dut** to some intrinsic difference between both groups

Some broad thoughts...

* In these posts, can you write down the model equation? That would help a lot. 

* Create a tests set to estimate performance. 

* It would really be a good idea to rectify your terminology with the statistical literature. Event time analysis really originated in the statistical literature and some of your terminology would sound pretty outlandish to anyone with a statistical background. There are probably statistical analogs to quantities like the CCDF and such. The analysis isn't bad, it just lacks all of the things that we've learned in the last 100 years about event time analysis. The remaining comments below reflect this point.

* I would avoid using `lm()`; nobody would use it that way. For event times, it is pretty common to use a log link for the outcome (even if the distribution is only defined for positive values).  I have code below for you to compare to `survreg` and the Bayesian analog. 

  * As a corollary to the above, you should use a log link in your analysis. It is accepted that event time models have a multiplicative error structure. 

* Also, I would use the concordance statistics to measure performance instead of RMSE. See `?survival::concordance` and the example code below. 

Example code:

``` r
library(tidymodels)
#> ── Attaching packages ──────────────────────────────────────────────────────────────────────────────────────────────────── tidymodels 0.0.2 ──
#> ✔ broom     0.5.1          ✔ purrr     0.3.2     
#> ✔ dials     0.0.2.9000     ✔ recipes   0.1.6.9000
#> ✔ dplyr     0.8.3          ✔ rsample   0.0.5     
#> ✔ ggplot2   3.2.0          ✔ tibble    2.1.3     
#> ✔ infer     0.4.0          ✔ yardstick 0.0.3.9000
#> ✔ parsnip   0.0.3
#> ── Conflicts ─────────────────────────────────────────────────────────────────────────────────────────────────────── tidymodels_conflicts() ──
#> ✖ purrr::discard()  masks scales::discard()
#> ✖ dplyr::filter()   masks stats::filter()
#> ✖ dplyr::lag()      masks stats::lag()
#> ✖ ggplot2::margin() masks dials::margin()
#> ✖ recipes::step()   masks stats::step()
library(survival)
library(brms)
#> Loading required package: Rcpp
#> 
#> Attaching package: 'Rcpp'
#> The following object is masked from 'package:rsample':
#> 
#>     populate
#> Loading 'brms' package (version 2.9.0). Useful instructions
#> can be found by typing help('brms'). A more detailed introduction
#> to the package is available through vignette('brms_overview').
#> 
#> Attaching package: 'brms'
#> The following object is masked from 'package:survival':
#> 
#>     kidney
#> The following object is masked from 'package:dials':
#> 
#>     mixture

set.seed(464)
tr_te_split <- initial_split(check_times, strata = "status")
check_time_tr <- training(tr_te_split)
check_time_te <- testing(tr_te_split)
```

``` r

sr_fit <-
  surv_reg(dist = "exponential") %>% 
  set_engine("survival") %>% 
  fit(Surv(check_time, status) ~ depends + imports + doc_size + r_size + 
        ns_import + ns_export, 
      data = check_time_tr)
tidy(sr_fit)
#> # A tibble: 7 x 7
#>   term         estimate std.error statistic  p.value conf.low conf.high
#>   <chr>           <dbl>     <dbl>     <dbl>    <dbl>    <dbl>     <dbl>
#> 1 (Intercept)  4.05      0.0173     233.    0.             NA        NA
#> 2 depends      0.111     0.00703     15.8   4.70e-56       NA        NA
#> 3 imports      0.0681    0.00333     20.4   1.11e-92       NA        NA
#> 4 doc_size     7.35      0.529       13.9   7.45e-44       NA        NA
#> 5 r_size       0.905     0.0935       9.68  3.72e-22       NA        NA
#> 6 ns_import    0.00450   0.00104      4.33  1.47e- 5       NA        NA
#> 7 ns_export   -0.000347  0.000402    -0.864 3.87e- 1       NA        NA

sr_pred <- 
  predict(sr_fit, check_time_te) %>% 
  bind_cols(check_time_te %>% select(check_time, status))  

concordance(Surv(check_time, status) ~ .pred, data = sr_pred)
#> Call:
#> concordance.formula(object = Surv(check_time, status) ~ .pred, 
#>     data = sr_pred)
#> 
#> n= 3406 
#> Concordance= 0.7754 se= 0.003941
#> concordant discordant     tied.x     tied.y    tied.xy 
#>    4439713    1285983         62      35214          1
```

``` r

# Recode the censoring indicators since brms does it completely differently
# From ?Surv: The status indicator, normally 0=alive, 1=dead.
# From ?brms::`addition-terms`: resp_cens requires 'left', 'none', 'right', and 
# 'interval' (or equivalently -1, 0, 1, and 2) to indicate left, no, right, or 
# interval censoring.

check_times_brm <- 
  check_time_tr %>% 
  mutate(status = ifelse(status == 1, 0, 1))

sink(tempfile()) # Stop the 10^28 lines of output generated by rstan::sampling
set.seed(355)
brm_fit <- 
  brm(check_time | cens(status) ~ depends + imports + doc_size + r_size + 
        ns_import + ns_export,
      data = check_times_brm, 
      family = exponential, 
      init = "0", 
      seed = 2593,
      cores = 4, 
      iter = 5000)
#> Compiling the C++ model
#> Start sampling
sink()

tidy(brm_fit) %>% as_tibble()
#> # A tibble: 8 x 5
#>   term             estimate std.error         lower         upper
#>   <chr>               <dbl>     <dbl>         <dbl>         <dbl>
#> 1 b_Intercept      4.05      0.0173        4.02          4.07    
#> 2 b_depends        0.111     0.00707       0.0993        0.123   
#> 3 b_imports        0.0680    0.00336       0.0625        0.0735  
#> 4 b_doc_size       7.37      0.524         6.52          8.22    
#> 5 b_r_size         0.906     0.0941        0.755         1.06    
#> 6 b_ns_import      0.00452   0.00105       0.00281       0.00624 
#> 7 b_ns_export     -0.000323  0.000400     -0.000964      0.000344
#> 8 lp__        -56588.        1.85     -56592.       -56586.
```

<sup>Created on 2019-07-29 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

topepo:
@kevinykuo 

> * if it errors right away for an unknown reason we also treat it as censored.

That's a whole different definition than the context here. We aren't looking at failure modes, just event times.

skeydan:
Thanks a lot Kevin!

>   When I see "censored" in the context of check times the first concept that comes to mind is "the check eventually completes, but it timed out so we don't know how long it actually takes." It seems like this isn't what's being considered, i.e. if it errors right away for an unknown reason we also treat it as censored. Might be worth a couple sentences to clarify this (though I might be the only one who had trouble with this.)

I see, makes sense! I'll add a footnote clarifying this.

> The quantiles are helpful, but it'd be useful to include a couple histogram plots I think as part of the "EDA".

Histograms are what I did first, but I found them to be less helpful, partly because of the big difference in sample size and partly because of fiddling with bin sizes ... In this case, at least to me, the numbers are a lot more expressive.

> Maybe expand on how we're using lm() for variable selection so we're not implicitly endorsing variable selection by p-values.

Makes sense, the main idea was to find an interesting, and useful, subset of predictors.

>  Should we compare to a surv_reg() instead of lm()? --> later ....

_posts/2019-07-30-censored-data/censored_data_tfp.Rmd
```diff
@@ -0,0 +1,724 @@
...
+
+In this post, we use a slightly less emotional example which nonetheless may be of interest, especially to R package developers: time to completion of `R CMD check`, collected from CRAN and provided by the `parsnip` package as `check_times`. Here, the censored portion are those checks that errored out for whatever reason, i.e., for which the check did not complete.
+
+Why do we care about the censored portion? In the cat adoption scenario, this is pretty obvious: We want to be able to get a realistic estimate for any unknown cat, not just those cats that will turn out to be "lucky". How about `check_times`? Well, if your submission is one of those that errored out, you still care about how long you wait, so even though their percentage is low (< 1%) we don't want to simply exclude them. Also, there is the possibility that the failing ones would have taken longer, had they run to completion, dut to some intrinsic difference between both groups. Conversely, if failures were random, the longer-running checks would have a greater chance to get hit by an error. So here too, exluding the censored data may result in bias.
```
javierluraschi: `dut to some ` -> `due to some`

skeydan: thanks!

skeydan:
Hi @topepo thanks for reviewing, and thanks for providing the code!

I'v e switched to using `surv_reg` throughout. However, I've not found a way to fit a model that incorporates that additional log link, even after extensive experimentation. As this model seems to do what it should, I think I'd like to publish it as is.

(In case alternative or better ways come up later, I can always comment on the post, or even write a follow-up.  To not stress the point, I'm not adding the equations now, but I'll keep that in mind for future posts. (((I always strive to explain everything as "verbally" as I can but I understand that for these models, the LaTeX really adds information and isn't just a necessity to put a paper on arxiv ;-)))

For the predictions, I think the `concordance` check is a bit out of scope, as this is not an introduction to survival analysis and suddenly I'd be presupposing a whole lot of unmentioned background. I've thought it a good compromise to just graphically display predictions from both models (on the test set) and not use any goodness of fit metric.

I think like this it should be good to publish.

skeydan:
after some more experimentation, there was a way to make it work with the log link - post updated :-)

PR #66, opened 2019-12-19
ready for review

skeydan:


_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
+---
+title: "Differential Privacy with TensorFlow"
+description: > 
+ Differential Privacy guarantees that results obtained from querying a database are basically independent of the presence in the data of a single individual. Applied to machine learning, we expect that no single training example influences the parameters of the trained model in a substantial way. This post introduces TensorFlow Privacy, a library built on top of TensorFlow, that can be used to train differentially private deep learning models from R.
```
batpigandme: Could the first part of that first sentence work as "the results of a database query..."?
I'm not sure how much is mitigating language by choice vs. by definition.

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+
+
+```{r setup, include=FALSE}
+knitr::opts_chunk$set(echo = TRUE, eval = FALSE)
+```
+
+
+What could be treacherous about summary statistics? 
+
+The famous _cat overweight_ study (X. et al., 2019) showed that as of May 1st, 2019, 32 of 101 domestic cats held in Y., a cozy Bavarian village, were overweight. Even though I'd be curious to determine if my aunt G.'s cat (a happy resident of that village) has been fed too many treats and has accumulated some excess pounds, from the results reported in that study I can't know.
```
batpigandme: > from the results reported in that study I can't know.

Maybe flip this line? "I can't determine that based on the results of the study."
(And then I suggest you change the earlier determine to "know" so it's not repetitive—also, I like having the informal language with your curiosity, and the more formal with the part referencing the study. 🙂 )

batpigandme:
```suggestion
The famous _cat overweight_ study (X. et al., 2019) showed that as of May 1st, 2019, 32 of 101 domestic cats held in Y., a cozy Bavarian village, were overweight. Even though I'd be curious to know if my aunt G.'s cat (a happy resident of that village) has been fed too many treats and has accumulated some excess pounds, this can't be determined based on the results of the study.
```

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+
+Then, six months later, out comes a new study, ambitious to earn scientific fame. The authors report that of 100 cats living in Y., 50 are striped, 31 are black, and the rest are white; the 31 black ones are all overweight. Now, I happen to know that with one exception, no new cat joined the community, and no cat left. _But_, my aunt moved away to town, to a retirement home, chosen of course for the possibility to bring one's cat.
```
batpigandme:
```suggestion
Then, six months later, out comes a new study, ambitious to earn scientific fame. The authors report that of 100 cats living in Y., 50 are striped, 31 are black, and the rest are white; the 31 black ones are all overweight. Now, I happen to know that, with one exception, no new cats joined the community, and no cats left. _But_, my aunt moved away from town to a retirement home, chosen of course for the possibility to bring one's cat.
```

batpigandme: Not sure why, but we tend to use the plural when speaking of hypothetical quantities. 

Is your aunt in the same town in this scenario? If yes, I'd simplify that last line to:
"_But_, my aunt moved to a retirement home, chosen of course for the possibility to bring one's cat.

batpigandme: ^^ I now understand that the answer to my question above is no — took me a second. (>ლ) I'm changing "to town" to "from town" since it makes it less ambiguous.

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+
+What have I just learned? My aunt's cat is overweight. (Or was, at least, before they moved to the retirement home.)
+
+Even though none of the studies reported anything but summary statistics, I was able to infer individual-level facts, by connecting both studies and adding in another piece of information I had access to.
```
batpigandme:
```suggestion
Even though none of the studies reported anything but summary statistics, I was able to infer individual-level facts by connecting both studies and adding in another piece of information I had access to. 
```

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+
+Even though none of the studies reported anything but summary statistics, I was able to infer individual-level facts, by connecting both studies and adding in another piece of information I had access to. 
+
+In reality, mechanisms like the above -- technically called _linkage_ -- have been shown to lead to privacy breaches many times, thus defeating the purpose of _database anonymization_ seen as a panacea in many organizations. A more promising alternative is offered by the concept of _differential privacy_.
+
+## Differential Privacy
+
+In differential privacy (DP)^[We'll be using DP as an acronym for both the noun phrase "differential privacy" and the adjective phrase "differentially private".][@Dwork2006], privacy is not a property of what's in the database; it's a property of how query results are delivered. 
+
+Intuitively paraphrasing results from a domain where results are communicated as theorems and proofs [@dwork2006differential][@Dwork], the only achievable (in a lossy but quantifiable way) objective is that from queries to a database, nothing more should be learned about an individual in that database than if they hadn't been in there at all.[@primer]
```
batpigandme:
```suggestion
Paraphrasing results from a domain where results are communicated as theorems and proofs [@dwork2006differential][@Dwork], the only achievable (in a lossy but quantifiable way) objective is that from queries to a database, nothing more should be learned about an individual in that database than if they hadn't been in there at all.[@primer] 
```

batpigandme: I don't think intuitively adds to what you're saying here, so I killed it. 😬

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+
+## Differential Privacy
+
+In differential privacy (DP)^[We'll be using DP as an acronym for both the noun phrase "differential privacy" and the adjective phrase "differentially private".][@Dwork2006], privacy is not a property of what's in the database; it's a property of how query results are delivered. 
+
+Intuitively paraphrasing results from a domain where results are communicated as theorems and proofs [@dwork2006differential][@Dwork], the only achievable (in a lossy but quantifiable way) objective is that from queries to a database, nothing more should be learned about an individual in that database than if they hadn't been in there at all.[@primer] 
+
+What this statement does is caution against overly high expectations: Even if query results are reported in a DP way (we'll see how that goes in a second), they enable some probabilistic inferences about individuals in the respective population. (Otherwise, why conduct studies at all.)
```
batpigandme:
```suggestion
What this statement does is caution against overly high expectations: Even if query results are reported in a DP way (we'll see how that goes in a second), they enable some probabilistic inferences about individuals in the respective population. (Otherwise, why conduct studies at all?)
```

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+
+This is reflected in the formal definition of DP. The idea is that queries to two databases differing in at most one element should give basically the same result. Put formally [@dwork2006differential]:
+
+> A randomized function $\mathcal{K}$ gives $\epsilon$ -differential privacy if for all data sets D1 and D2 differing on at most one element, and all $S \subseteq Range(K)$, 
+
+> $Pr[\mathcal{K}(D1)\in S] \leq exp(\epsilon) × Pr[K(D2) \in S]$
+
+This $\epsilon$ -differential privacy is additive: If one query is $\epsilon$-DP at a value of 0.01, and another one at 0.03, together they will be 0.04 $\epsilon$-differentially private.
+
+If $\epsilon$-DP is to be achieved via adding noise, how exactly should this be done? Here, several mechanisms exist; the basic, intuitively plausible principle though is that the amount of noise should be calibrated to the target function's sensitivity, defined as the maximum $l1$ norm of the difference of function values computed on all pairs of datasets differing in a single example.
```
batpigandme:
```suggestion
If $\epsilon$-DP is to be achieved via adding noise, how exactly should this be done? Here, several mechanisms exist; the basic, intuitively plausible principle, though, is that the amount of noise should be calibrated to the target function's sensitivity, defined as the maximum $l1$ norm of the difference of function values computed on all pairs of datasets differing in a single example.
```

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+
+So far, we've been talking about databases and datasets. How does this apply to machine and/or deep learning? 
+
+
+
+## TensorFlow Privacy
+
+Applying DP to deep learning, we want a model's parameters to wind up "essentially the same" whether trained on a dataset including that cute little kitty or not. TensorFlow (TF) Privacy [@abaditfp], a library built on top of TF, makes it easy on users to add privacy guarantees to their models -- easy, that is, from a technical point of view. (As with life overall, the hard decisions on _how much_ of an asset we should be reaching for, and how to trade off one asset (here: privacy) with another (here: model performance), remain to be taken by each of us ourselves.)
```
batpigandme:
```suggestion
Applying DP to deep learning, we want a model's parameters to wind up "essentially the same" whether trained on a dataset including that cute little kitty or not. TensorFlow (TF) Privacy [@abaditfp], a library built on top of TF, makes it easy for users to add privacy guarantees to their models -- easy, that is, from a technical point of view. (As with life overall, the hard decisions on _how much_ of an asset we should be reaching for, and how to trade off one asset (here: privacy) with another (here: model performance), remain to be taken by each of us ourselves.)
```

batpigandme: Changed "easy on users" to "easy for users" (IDK why it doesn't show up in the diff).

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+# EKG data - provided in already averaged form, at frequency 0.5 Hz
+s1$label %>% nrow() # 4603
+```
+
+In light of the different sampling frequencies, our `tfdatasets` pipeline will have do some moving averaging, paralleling that applied to construct the ground truth data.
+
+
+## Preprocessing pipeline
+
+As every "column" ^[For convenience, we take the liberty to talk as if we had the usual rectangular data here.] is of different length and resolution, we build up the final _dataset_ piece-by-piece.
```
batpigandme:
```suggestion
As each "column" ^[For convenience, we take the liberty to talk as if we had the usual rectangular data here.] is of a different length and resolution, we build up the final _dataset_ piece-by-piece.
```

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+n_total <- 4603 # keep track of this
+
+# keep matching numbers of observations of predictors
+acc <- s1$signal$wrist$ACC[1:(n_total * 64), ] # 32 Hz, 3 columns
+bvp <- s1$signal$wrist$BVP[1:(n_total * 128)] %>% matrix() # 64 Hz
+eda <- s1$signal$wrist$EDA[1:(n_total * 8)] %>% matrix() # 4 Hz
+temp <- s1$signal$wrist$TEMP[1:(n_total * 8)] %>% matrix() # 4 Hz
+```
+
+Some more housekeeping. Both training and the test set need to have a `timesteps` dimension, as usual with architectures that work on sequential data (1-d convnets and RNNs). To make sure there is no overlap between respective `timesteps`, we split the data "up front" and assemble both sets separately. We'll use the first 4000 observations for training.
```
batpigandme:
```suggestion
Some more housekeeping: Both the training and the test sets need to have a `timesteps` dimension, as usual with architectures that work on sequential data (1-d convnets and RNNs). To make sure there is no overlap between respective `timesteps`, we split the data "up front" and assemble both sets separately. We'll use the first 4000 observations for training.
```

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+
+ds_test <- ds_test %>%
+  # see above reg. dataset_repeat
+  dataset_repeat() %>%
+  dataset_batch(batch_size)
+
+
+```
+
+With data manipulations as complicated as the above, it's always worthwhile checking some pipeline outputs. We can do that using the usual `reticulate::as_iterator` magic, provided that for this test run, we _don't_ disable V2 behavior. (Just restart the R session between a "pipeline checking" and the later modeling runs.)
```
batpigandme:
```suggestion
With data manipulations as complicated as the above, it's always worthwhile to check some pipeline outputs. We can do that using the usual `reticulate::as_iterator` magic, provided that for this test run, we _don't_ disable TF V2 behavior. (Just restart the R session between a "pipeline checking" and the later modeling runs.)
```

batpigandme: Added the TF, since there's plenty the reader's holding in their brains between the previous discussion of it and now!

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+iter <- as_iterator(ds_test) # or any other dataset you want to check
+while (TRUE) {
+ item <- iter_next(iter)
+ if (is.null(item)) break
+ print(item)
+}
+```
+
+
+With that we're ready to create the model.
```
batpigandme:
```suggestion
With that, we're ready to create the model.
```

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+
+## Model
+
+The model will be a rather simple convnet. The main difference between standard and DP training lies in the optimization procedure; thus, it's straightforward to first establish a non-DP baseline. Later, when switching to DP, we'll be able to reuse almost everything.
```
batpigandme:
```suggestion
The model will be a rather simple convnet. The main difference between standard and DP training lies in the optimization procedure. Thus, it's straightforward to first establish a non-DP baseline. Later, when switching to DP, we'll be able to reuse almost everything.
```

batpigandme: I love a good semicolon as much as anyone, but I think the first part of that sentence is sufficiently important that it should stand on its own.

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+## Baseline results
+
+
+After 20 epochs, mean absolute error is around 6 bpm: 
+
+```{r, eval=TRUE, echo=FALSE, layout="l-body-outset", fig.cap = "Training history without differential privacy."}
+knitr::include_graphics("images/baseline.png")
+```
+
+Just to put this in context, the MAE reported for subject S1 in the paper[@ReissISL19] -- based on a higher-capacity network, extensive hyperparameter tuning, and naturally, training on the complete dataset -- amounts to 8.45 bpm on average; so our setup seems to be sound.
```
batpigandme:
```suggestion
To put this in context, the MAE reported for subject S1 in the paper[@ReissISL19] -- based on a higher-capacity network, extensive hyperparameter tuning, and naturally, training on the complete dataset -- amounts to 8.45 bpm on average. So, our setup seems to be sound.
```

batpigandme: I've been trying to get rid of "just"s in my writing of late (turns out I use it _a lot_).

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+
+Now we'll make this differentially private.
+
+
+## DP training
+
+Instead of plain `Adam`, we use the corresponding TF Privacy wrapper, `DPAdamGaussianOptimizer`.
```
batpigandme: I had to <kbd>Ctrl</kbd> + <kbd>f</kbd> to figure out who Adam was 😂! I'm guessing this is something most readers will be familiar with, but it couldn't hurt to add "the plain `Adam` optimizer," or something of the like.

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+
+We need to tell it how aggressive gradient clipping should be (`l2_norm_clip`) and how much noise to add (`noise_multiplier`). Furthermore, we define the learning rate (there is no default), going for 10 times the default `0.001` based on initial experiments.
+
+There is an additional parameter, `num_microbatches`, that could be used to speed up training [@abs181206210], but as training duration is not an issue here we just set it equal to `batch_size`.
```
batpigandme:
```suggestion
There is an additional parameter, `num_microbatches`, that could be used to speed up training [@abs181206210], but, as training duration is not an issue here, we just set it equal to `batch_size`.
```

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+DP-SGD with sampling rate = 0.802% and noise_multiplier = 1.1 iterated over
+2494 steps satisfies differential privacy with eps = 2.73 and delta = 1e-06.
+```
+
+How good is a value of 2.73? Citing the [TF Privacy authors](https://github.com/tensorflow/privacy/tree/master/tutorials):
+
+> $\epsilon$ gives a ceiling on how much the probability of a particular output can increase by including (or removing) a single training example. We usually want it to be a small constant (less than 10, or, for more stringent privacy guarantees, less than 1). However, this is only an upper bound, and a large value of epsilon may still mean good practical privacy.
+
+
+Obviously, choice of $\epsilon$ is a (challenging) topic by itself, and not something we can elaborate on in a post dedicated to the technical aspects of DP with TensorFlow.
```
batpigandme:
```suggestion
Obviously, choice of $\epsilon$ is a (challenging) topic unto itself, and not something we can elaborate on in a post dedicated to the technical aspects of DP with TensorFlow.
```

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+optimizer <- priv$DPAdamGaussianOptimizer(
+  l2_norm_clip = l2_norm_clip,
+  noise_multiplier = noise_multiplier,
+  num_microbatches = num_microbatches,
+  learning_rate = learning_rate
+)
+
+```
+
+One other change to make for DP. As gradients are clipped on a per-sample basis, the optimizer needs to work with per-sample losses as well:
```
batpigandme:
```suggestion
There is one other change to make for DP. As gradients are clipped on a per-sample basis, the optimizer needs to work with per-sample losses as well: 
```

batpigandme: That first sentence was a fragment, the other way to go would be to use the original wording and end it with a colon. But, given the following sentence ends with a colon, I think it's easier to turn that first line into a sentence of its own.

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+priv$privacy$analysis$rdp_accountant$get_privacy_spent(
+  orders, rdp, target_delta = 1e-6)[[1]]
+
+```
+
+```
+[1] 4.249645
+```
+
+So, we do get the same result. This was a long post; let's wrap up.
```
batpigandme:
```suggestion
So, we do get the same result.
```

batpigandme: I like having that conclusion on its own, since it's quite a journey to get there. I think the `## Conclusion` header is sufficient for signalling that you're wrapping up. 😸

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
skeydan: oh i see, if the language makes you think there's something wrong... I'll just remove that town thing completely, retirement home should be enough :-)

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
skeydan: "unto itself" ... need to keep that in mind (wasn't aware of)

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
skeydan: Hm, we had the same situation in the last post... Is it too informal (I know I'm always on the verge ;-)) or is just totally uncommon in English?

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
skeydan: ooooooh yeah... among every thing else, these are the kinds of comments I love your reviews for :-))) (take out the fill words / fill phrases or however that's called ... the no-information stuff.)

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
skeydan: true dat ;-)))

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
skeydan: oh, sure! will change

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
skeydan: makes sense! will change

skeydan: actually, found  yet sth else: "the study results don't tell."

skeydan:
thanks a lot Mara, super helpful as always!!

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+
+library(purrr)
+
+library(reticulate)
+# if you haven't yet, install TF Privacy, e.g. using reticulate:
+# py_install("tensorflow_privacy")
+priv <- import("tensorflow_privacy")
+```
+
+From the downloaded archive, we just need `S1.pkl`, saved in a [native Python serialization format](https://docs.python.org/3.8/library/pickle.html), yet nicely loadable using `reticulate`:
```
kevinykuo: Understanding that the focus of this blog post is DP and not data ingest... is there a CSV version of the data? If it's not too much work to start from a CSV I think you'd be able to relate to a lot more R users.

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+As every "column" ^[For convenience, we take the liberty to talk as if we had the usual rectangular data here.] is of different length and resolution, we build up the final _dataset_ piece-by-piece.
+The following function serves two purposes:
+
+1. compute running averages over differently sized windows, thus downsampling to 0.5Hz for every modality
+2. transform the data to the `(num_timesteps, num_features)` format that will be required by the 1d-convnet we're going to use soon
+
+```{r}
+average_and_make_sequences <-
+  function(data, window_size_avg, num_timesteps) {
+    data %>% tf$cast(tf$float32) %>%
```
kevinykuo: Consider one pipe per line

kevinykuo: Somewhat off-topic, do you think it's acceptable style to use `k_` functions where available to avoid `$`? E.g. `k_cast()` instead of `tf$cast()`.

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+n_total <- 4603 # keep track of this
+
+# keep matching numbers of observations of predictors
+acc <- s1$signal$wrist$ACC[1:(n_total * 64), ] # 32 Hz, 3 columns
+bvp <- s1$signal$wrist$BVP[1:(n_total * 128)] %>% matrix() # 64 Hz
+eda <- s1$signal$wrist$EDA[1:(n_total * 8)] %>% matrix() # 4 Hz
+temp <- s1$signal$wrist$TEMP[1:(n_total * 8)] %>% matrix() # 4 Hz
+```
+
+Some more housekeeping. Both training and the test set need to have a `timesteps` dimension, as usual with architectures that work on sequential data (1-d convnets and RNNs). To make sure there is no overlap between respective `timesteps`, we split the data "up front" and assemble both sets separately. We'll use the first 4000 observations for training.
```
kevinykuo: I think a line chart somewhere of some of these sequential predictors would be helpful in getting oriented.

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+
+```
+
+With data manipulations as complicated as the above, it's always worthwhile checking some pipeline outputs. We can do that using the usual `reticulate::as_iterator` magic, provided that for this test run, we _don't_ disable V2 behavior. (Just restart the R session between a "pipeline checking" and the later modeling runs.)
+
+Here, in any case, would be the relevant code:
+
+
+```{r}
+# this piece needs TF 2 behavior enabled
```
kevinykuo: What's the incantation to enable this behavior?

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
```diff
@@ -0,0 +1,470 @@
...
+
+This is reflected in the formal definition of DP. The idea is that queries to two databases differing in at most one element should give basically the same result. Put formally [@dwork2006differential]:
+
+> A randomized function $\mathcal{K}$ gives $\epsilon$ -differential privacy if for all data sets D1 and D2 differing on at most one element, and all $S \subseteq Range(K)$, 
+
+> $Pr[\mathcal{K}(D1)\in S] \leq exp(\epsilon) × Pr[K(D2) \in S]$
+
+This $\epsilon$ -differential privacy is additive: If one query is $\epsilon$-DP at a value of 0.01, and another one at 0.03, together they will be 0.04 $\epsilon$-differentially private.
+
+If $\epsilon$-DP is to be achieved via adding noise, how exactly should this be done? Here, several mechanisms exist; the basic, intuitively plausible principle though is that the amount of noise should be calibrated to the target function's sensitivity, defined as the maximum $l1$ norm of the difference of function values computed on all pairs of datasets differing in a single example.
```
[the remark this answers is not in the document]
skeydan: makes sense, adding :-)

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
skeydan: I see, but on the other hand:

- It also makes sense to show we can handle inconvenient file formats found "in the wild"
- we want to show tfdatasets pipelines as _the_ way to do preprocessing for DL, if possible
- especially for that topic of transforming to `timesteps`, I think that once one has understood how to do this with `dataset_window` it's actually the easiest way (ok, rephrasing: once one has an example to copy-paste :-)) - myself, for example, I'll definitely stick to this way now that I've found out ;-))

BUT, I do think you might be writing this having some other project in mind. There, it will definitely be a thing to consider, and we should discuss the target group thing in more detail next month.

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
skeydan: right, changing...

skeydan: yeah you may be right, it probably is preferred style-wise (?)
I think I got "spoiled" by the TFP applications not involving `keras`, so I tend to automatically use `tf$` style sometimes...
Changing.

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
skeydan: I think I'll skip this one, as probably the dataset related content is on the verge of distracting too much from the main topic already...

_posts/2019-12-20-differential-privacy/differential_privacy.Rmd
skeydan: right, I'll make this more explicit

skeydan:
thanks a lot Kevin, very helpful!

PR #72, opened 2020-03-27
explanatory post accompanying name change

skeydan:
TBD:
- actually change title
- edit static pages where adequate
- consolidate categories in order to have everything on a less fine-grained level (and eliminate organically-grown redundancies)

All done! Apart from changing `baseurl` in site.yaml.


javierluraschi:
LGTM!

Note to future self: We have a follow up TODO to rename the URL from `/tensorflow` to `/ai` in https://github.com/rstudio/blogs.rstudio.com/blob/master/build.sh#L29; however, I think changing the title/content first and then placing the redirect should be better for SEO ranking and overall less disruptive, we will follow up on that.!

PR #108, opened 2020-09-29
fix install instructions to match collab notebook

daviddalpiaz:
It seems that `torchvision` is not on CRAN, so following the post instructions is not reproducible.


skeydan:
thanks!

daviddalpiaz:
Ah, seems the change isn't live because I didn't re-build the project! Sorry. I didn't want to push my changes to the `.html` because it seemed to change a lot of it and I was worried I was running an old version of something somewhere so my render wouldn't be what was expected.

skeydan:
don't worry - I'll  rebuild it for tomorrow :-)

and I totally agree about being cautious :-)

daviddalpiaz:
:)

PR #147, opened 2021-03-12
add seq2seq

skeydan:


_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
```diff
@@ -0,0 +1,445 @@
...
+
+Quite fittingly, the types of models employed to these ends are named sequence-to-sequence models (often abbreviated *seq2seq*). In a nutshell, they split up the task into two components: an encoding and a decoding part. The former is done just once per input-target pair. The latter is done in a loop, as in our first try. But the decoder has more information at its disposal: At each iteration, its processing is based on the previous prediction as well as previous state. That previous state will be the encoder's when a loop is started, and its own ever thereafter.
+
+Before discussing the model in detail, we need to adapt our data input mechanism.
+
+# Data input
+
+We continue working with `vic_elec` , provided by `tsibbledata`.
+
+Again, the dataset definition in the current post looks a bit different from the way it did before; again, it's the shape of the target that differs. This time, `y` equals `x`, shifted to the left by one.
```
batpigandme, line 44:
```suggestion
Again, the dataset definition in the current post looks a bit different from the way it did before; it's the shape of the target that differs. This time, `y` equals `x`, shifted to the left by one.
```

_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
```diff
@@ -0,0 +1,445 @@
...
+
+The reason we do this is owed to the way we are going to train the network. With *seq2seq*, people often use a technique called "teacher forcing", where instead of feeding back its own prediction into the decoder module, you pass it the value it *should* have predicted. To be clear, this is done during training only, and to a configurable degree.
```
batpigandme, line 46:
```suggestion
The reason we do this is owed to the way we are going to train the network. With *seq2seq*, people often use a technique called "teacher forcing" where, instead of feeding back its own prediction into the decoder module, you pass it the value it *should* have predicted. To be clear, this is done during training only, and to a configurable degree.
```

_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
```diff
@@ -0,0 +1,445 @@
...
+test_dl <- test_ds %>% dataloader(batch_size = 1)
+```
+
+# Model
+
+Technically, the model consists of three *modules*: the aforementioned encoder and decoder, and the *seq2seq* module that orchestrates them.
+
+### Encoder
+
+The encoder takes its input and runs it through an RNN. Of the two things returned by a recurrent neural network, outputs and state, so far we've only been using the former. This time, we do the opposite: We throw away the outputs, and only return the state.
```
batpigandme, line 123:
```suggestion
The encoder takes its input and runs it through an RNN. Of the two things returned by a recurrent neural network, outputs and state, so far we've only been using output. This time, we do the opposite: We throw away the outputs, and only return the state.
```

_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
```diff
@@ -0,0 +1,445 @@
...
+
+If the RNN in question is a GRU (and assuming that of the outputs, we take just the final time step, which is what we've been doing throughout), there really is no difference: The final state equals the final output. If it's an LSTM, however, there is a second, the "cell", state. In that case, returning the state instead of the final output will carry more information.
```
batpigandme, line 125:
```suggestion
If the RNN in question is a GRU (and assuming that of the outputs, we take just the final time step, which is what we've been doing throughout), there really is no difference. The final state equals the final output. If it's an LSTM, however, there is a second state, the "cell". In that case, returning the state instead of the final output will carry more information.
```

batpigandme: You've got a lot of the colon-to-new-sentence construct, so I'm paring that back a bit. Also, unless there are two "the cells," then the second sentence I modified doesn't make sense as is (though it should with the edit).

_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
```diff
@@ -0,0 +1,445 @@
...
+    x
+    
+  }
+  
+)
+```
+
+### Decoder
+
+In the decoder, too, the main component is an RNN. In contrast to previously-shown architectures, though, it does not just return a prediction. It also reports back the RNN's final state.
```
batpigandme, line 170:
```suggestion
In the decoder, the main component is an RNN. In contrast to previously-shown architectures, though, it does not just return a prediction. It also reports back the RNN's final state.
```

_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
```diff
@@ -0,0 +1,445 @@
...
+    list(output, next_hidden)
+    
+  }
+  
+)
+```
+
+### `seq2seq` module
+
+In `seq2seq` is where the action happens. The plan is to encode once, then call the decoder in a loop.
```
batpigandme, line 223:
```suggestion
`seq2seq` is where the action happens. The plan is to encode once, then call the decoder in a loop.
```

_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
```diff
@@ -0,0 +1,445 @@
...
+
+If you look back to decoder `forward()`, you see that it takes two arguments: `x` and `state`.
+
+Dependent on the context, `x` corresponds to one of three things.
```
batpigandme, line 227:
```suggestion
Depending on the context, `x` corresponds to one of three things.
```

batpigandme: What are the three things? If they're listed below, it's a bit unclear what the exact correspondence is, so you might give abbreviated versions of them here.

_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
```diff
@@ -0,0 +1,445 @@
...
+
+-   The very first time the decoder is called on an input sequence, `x` maps to the final input value. This is different from a task like machine translation, where you would pass in a start token. With time series, though, we'd like to continue where the actual measurements stop.
+
+-   In further calls, we want the decoder to continue from its most recent prediction. It is only logical, thus, to pass back the preceding forecast.
```
batpigandme, line 231: So, does x map to the preceding forecast?

_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
```diff
@@ -0,0 +1,445 @@
...
+
+-   That said, in NLP a technique called "teacher forcing" is commonly used to speed up training. With teacher forcing, instead of the forecast we pass the actual ground truth, the thing the decoder should have predicted. We do that only in a configurable fraction of cases, and -- naturally -- only while training. The rationale behind this technique is that without this form of re-calibration, consecutive prediction errors can quickly erase any remaining signal.
```
batpigandme, line 233: You're bordering on this not really being a bullet item here. If you're leaving these as bullets, be sure to have the "answer" (the item x is mapping to) most prominent or in the same place every time. Right now, it can get lost in the caveats and explanations.

_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
```diff
@@ -0,0 +1,445 @@
...
+
+`state`, too, is polyvalent. But here, there are just two possibilities.
+
+-   The first time the decoder is called, it is "seeded" with the final state from the encoder. Note how this is *the only time* we make use of the encoding.
+
+-   From then on, the decoder's own previous state will be passed. Remember how it returns two values, forecast and state?
```
batpigandme, line 239: Again, you want what's called "list agreement" with your bullets. If you're giving us the two possibilities, give them to us first, and then expand on the idea.

_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
```diff
@@ -0,0 +1,445 @@
...
+# alternatively, use 
+# device <- "cpu"
+device <- torch_device(if (cuda_is_available()) "cuda" else "cpu")
+
+net <- net$to(device = device)
+```
+
+# Training
+
+The training procedure is *mainly* unchanged. We do, however, need to decide about `teacher_forcing_ratio`, the proportion of input sequences we want to perform re-calibration on. In `valid_batch`, this should always be `0`, while in `train_batch`, it's up to us (or rather, experimentation). Here, we set it to `0.3`.
```
batpigandme, line 306:
```suggestion
The training procedure is *mainly* unchanged. We do, however, need to decide about `teacher_forcing_ratio`, the proportion of input sequences we want to perform re-calibration on. In `valid_batch()`, this should always be `0`, while in `train_batch()`, it's up to us (or rather, experimentation). Here, we set it to `0.3`.
```

batpigandme: Functions get trailing parentheses to distinguish them from arguments.

_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
```diff
@@ -0,0 +1,445 @@
...
+  autoplot() +
+  scale_colour_manual(values = c("#08c5d1", "#00353f", "#ffbf66", "#d46f4d")) +
+  theme_minimal()
+```
+
+```{r, eval = TRUE, echo = FALSE, layout="l-page", fig.cap="One-week-ahead predictions for January, 2014."}
+knitr::include_graphics("images/seq2seq_preds.png")
+```
+
+Comparing this to the forecast obtained from last time's RNN-MLP combo, we don't see much of a difference. Is this surprising? To me it is. If asked to speculate about the reason, I would probably say that, likely, in all of the setups we've used so far, the main carrier of information has been the final hidden state of the RNN (the single RNN in the two previous architectures, the *encoder* RNN in this one). It will be interesting to check this hypothesis in the last part of this series, where we augment the encoder-decoder architecture by *attention*.
```
batpigandme, line 439: You've got five commas and a parenthetical in the sentence beginning “If asked to speculate…”, which is a sign that you should probably rework it. I'll leave that to you, but note that the first half:
> If asked to speculate about the reason, I would probably say that, likely…

is almost all setup for actual content.

_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
skeydan: hm, no, it really should be "cell state" ... I think I can do
> second kind of state, the "cell state"

batpigandme: Yeah, sounds much better. Originally there was a comma between "cell" and state, so I didn't even put them together.

_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
skeydan: ok, perhaps I should make more clear what the reference is, here ....as in
> In the decoder, just like in the encoder, the main component

_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
skeydan: oh, yes, I see... will mention them here already

_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
skeydan: yes - that should get more clear when I incorporate your previous suggestion

_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
skeydan: hm, hm... I see what you mean ... I do find it a bit "refreshing" tbh, as in, deviating a bit from expectations ...
... just like you express it, "bordering on" :-)

let me think about it again when I'm in the second stage (after the "commit-ready" ones) ...

_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
skeydan: forgot, sorry

_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
skeydan: haha - i see ... that was a comma of type "no idea whether it should go there or not" :-)

_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
skeydan: @batpigandme if by chance you see this - could you let me know if this is correct, especially, if this is a correct use of "leading"? I found "leading" as a synonym for "preceding", but have no idea if it can be used like that (actually, that sense of "leading" I was only familiar with from window function arguments - lag & lead :-))

> Depending on the context, `x` corresponds to one of three things: final input, preceding prediction, or leading ground truth.

_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
skeydan: yup. Going for a compromise :-) (keeping some "deviance" ;-))

>  But here, there are just two possibilities: encoder state and decoder state.

_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
skeydan: ooooh yeah ... you're right, in fact that sentence should never have made it to editor review :-)))

_posts/2021-03-15-forecasting-time-series-with-torch_3/forecasting_3.Rmd
batpigandme: For leading do you mean prior? I'm not 100% clear on what its intended meaning is here. 😬

skeydan: yes, same as preceding/previous/antecedent/....?

but now that you're suggesting it: **prior** might be best :-)

PR #154, opened 2021-03-25
implement caching for R packages and build artifacts for AI blog

yitao-li:
There are 2 major changes related to caching, and then 1 more change related to necessary cache invalidation of build artifacts:

- Required R packages for rendering are cached. Those packages are first fetched from GH action cache if possible, and are replaced with more up-to-date versions if newer versions are available. This type of caching usually saves about 7 minutes of build time.

- Implementing support for incremental build: build artifacts are first fetched from GH action cache if possible, and if the Rmd source file that generated a cached build artifacts is different from the one in current repo, then re-render the Rmd source file in the current repo. This type of caching usually saves vast majority of build time because it allows us to only render content that has changed rather than rendering all posts from scratch in each update. This means build time for each update decreases from 30+ minutes to about 2 minutes.

- However, if any of the following changes, then existing build artifacts cache must be invalidated and we must re-build everything from scratch and then cache the more up-to-date build:

  * R version used
  * Pandoc version used
  * Required list of R packages for rendering the posts
  * Latest versions of those R packages (or commit hash if installed from github)


dfalbel (commented):
Looks great to me!

I think we could avoid some complexity by using fixed versions of the R packages that we are using (specially distill and rmarkdown) and R itself instead of relying on the latest versions, but that's not mandatory and I think using the latest versions is also fine!

PR #158, opened 2021-04-26
add optimization post

skeydan:


_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
```diff
@@ -0,0 +1,640 @@
...
+preview: images/preview.jpg
+---
+
+```{r setup, include=FALSE}
+knitr::opts_chunk$set(echo = TRUE, eval = FALSE)
+```
+
+So far, all `torch` use cases we've seen have been in deep learning. However, its automatic differentiation feature is useful in other areas, as well. One prominent example is numerical optimization: We can use `torch` to find the minimum of a function.
+
+In fact, function minimization is *exactly* what happens in training a neural network. But there, the function in question normally is far too complex to even imagine finding its minima analytically. Numerical optimization aims at building up the tools to handle just this complexity. To that end, however, it starts from functions that are far less deeply composed. Instead, they are hand-crafted to pose specific challenges.
```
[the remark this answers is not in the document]
batpigandme: This paragraph is a little bit confusing to me, which I'm sure is at least in part due to my lack of technical understanding. Here's what I'm getting from it, and you can decide if any of this might be refactored:

* Function minimization is what happens when you're training a neural network, but that function is too complicated to imagine finding its minima.
* Numerical optimization is how we make tools to handle the unimaginably complex kind of function(?)
* To do this is starts with functions that are "less deeply composed" (so, I'm guessing easier to imagine? i.e. we can't imagine the ones from neural networks because they're super deeply composed?)

Summing up the last sentence but filling in the blanks/pronouns with what we've learned above:
* Numerical optimization starts from functions that are hand-crafted (by us?) to pose specific challenges (which are less complex, so we can imagine them)

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
```diff
@@ -0,0 +1,640 @@
...
+
+This post is a first introduction to numerical optimization with `torch`. Central takeaways are the existence and usefulness of its L-BFGS optimizer, as well as the impact of running L-BFGS with line search. As a fun add-on, we show an example of constrained optimization, where a constraint is enforced via a quadratic penalty function.
```
batpigandme, line 29: > This post is a first introduction to numerical optimization with `torch`.

This is probably worth having in the description to the post, since it's a concise, relevant reason as to why someone would read this.

batpigandme: **L-BFGS** is this a very well known initialism? If not, you should spell it out before using it. If so, you can probably wait until you define it, but, given distill allows for margin notes, etc., there's no real reason _not_ to give your reader the words behind a given acronym/initialism at _some_ point.

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
```diff
@@ -0,0 +1,640 @@
...
+
+To warm up, we take a detour, minimizing a function "ourselves" using nothing but tensors. This will turn out to be less of a detour than expected though: Later, the overall process will still be the same. All changes will be related to integration of `optimizer`s and their capabilities.
```
batpigandme, line 31:
```suggestion
To warm up, we take a detour, minimizing a function "ourselves" using nothing but tensors. This will turn out to be relevant later, though, as the overall process will still be the same. All changes will be related to integration of `optimizer`s and their capabilities.
```

batpigandme: What is `optimizer`? You use backticks around R package names in your posts. Backticks also go around code (e.g. functions and/or their arguments—though you should have trailing parentheses if it's a function).

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
```diff
@@ -0,0 +1,640 @@
...
+
+## Function minimization, DYI approach
+
+To see how we can minimize a function "by hand", let's try the iconic Rosenbrock function. This is a function of two variables:
```
batpigandme, line 35:
```suggestion
To see how we can minimize a function "by hand", let's try the iconic [Rosenbrock function](https://en.wikipedia.org/wiki/Rosenbrock_function). This is a function with two variables:
```

batpigandme: Even if I'm the only one to read this post who _doesn't_ know what the Rosenbrock function is, it can't hurt to link out to it.

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
```diff
@@ -0,0 +1,640 @@
...
+
+```{r, eval = TRUE, echo = FALSE, layout="l-page", fig.cap="Rosenbrock function."}
+knitr::include_graphics("images/rosenbrock.png")
+```
+
+Our goal and strategy are as follows.
+
+We want to find the values $x_1$ and $x_2$ for which the function attains its minimum. We have to start somewhere; and from wherever that gets us on the graph we follow the negative of the gradient "downwards", descending into regions of consecutively smaller function value.
+
+Concretely, in every iteration, we take the current $(x1,x2)$ point, compute the function value as well as the gradient, and subtract some fraction of the latter to arrive at a new $(x1,x2)$ candidate. This process goes on until we either reach the minimum -- the gradient is zero -- or improvement is below a chosen threshold.
```
batpigandme, line 70: 👍  good description

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
```diff
@@ -0,0 +1,640 @@
...
+
+    Iteration:  900 
+    Value is:  0.0001532408 
+    Gradient is:  -0.004811743 -0.009894371 
+
+    Iteration:  1000 
+    Value is:  6.962555e-05 
+    Gradient is:  -0.003222887 -0.006653666 
+
+While this works, it really serves to illustrate the principle only. With `torch` providing a bunch of proven optimization algorithms, there is no need for us to manually compute the candidate $\mathbf{x}$ values.
```
batpigandme, line 123:
```suggestion
While this works, it really serves to illustrate the principle. With `torch` providing a bunch of proven optimization algorithms, there is no need for us to manually compute the candidate $\mathbf{x}$ values.
```

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
```diff
@@ -0,0 +1,640 @@
...
+
+## Function minimization with `torch` optimizers
+
+Instead, we let a `torch` optimizer update the candidate $\mathbf{x}$ for us. Habitually, our first try is *Adam*.
+
+### Adam (?)
```
batpigandme, line 129: Is naming attempts a numerical-optimization/computing/technical thing? If not, at this point (I'm reading the post in order, obviously), I'm not sure why we're doing this.

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
```diff
@@ -0,0 +1,640 @@
...
+    Value is:  4.003079e-05 
+    Gradient is:  -0.05383469 0.02346456 
+
+    Iteration:  100 
+    Value is:  6.937736e-05 
+    Gradient is:  -0.003240437 -0.006630421 
+
+It took us about a hundred iterations to arrive at a decent value. This is a lot faster than the manual approach above, but still quite a lot. Luckily, further improvements are possible.
+
+### L-BFGS (!)
```
batpigandme, line 179: I now see that it's quite a mouthful to spell this one out. You could have it as a hoverable footnote earlier, perhaps?

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
```diff
@@ -0,0 +1,640 @@
...
+
+Among the many `torch` optimizers commonly used in deep learning (Adam, AdamW, RMSprop ...), there is one "outsider", much better known in classic numerical optimization than in neural-networks space: L-BFGS, a.k.a. [Limited-memory BFGS](https://en.wikipedia.org/wiki/Limited-memory_BFGS), a memory-optimized implementation of the [Broyden--Fletcher--Goldfarb--Shanno optimization algorithm](https://en.wikipedia.org/wiki/Broyden%E2%80%93Fletcher%E2%80%93Goldfarb%E2%80%93Shanno_algorithm) (BFGS).
```
batpigandme, line 181: I now see that Adam is an actual torch optimizer. Maybe it was the question mark after it in the subheader that threw me earlier. Your call as to whether or not readers will already know this. If not, maybe refer to it as "the Adam optimizer" at first, or something of the like.

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
```diff
@@ -0,0 +1,640 @@
...
+
+BFGS is perhaps the most widely used among the so-called Quasi-Newton, second-order optimization algorithms. As opposed to the family of first-order algorithms that, in deciding on a descent direction, make use of gradient information only, second-order algorithms additionally take curvature information into account. To that end, exact Newton methods actually compute the Hessian (a costly operation), while Quasi-Newton methods avoid that cost and, instead, resort to iterative approximation.
```
batpigandme, line 183: > As opposed to the family of first-order algorithms that, in deciding on a descent direction, make use of gradient information only, second-order algorithms additionally take curvature information into account.

This sentence makes sense when I get to the end of it, but it's almost as though the information is presented backwards (I don't know what you're describing—what information is taken into account with second-order algorithms—until the end).

Given this paragraph is defining technical terminology, try giving your reader simpler sentences that clearly define each component. You can even tell the reader that you're breaking down the definition/description given in that first sentence (_so-called Quasi-Newton, second-order optimization algorithm_). If someone already knew what you meant in that sentence, they'll move on. If someone (e.g. me) didn't know, they won't have to work so hard to figure it out.

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
```diff
@@ -0,0 +1,640 @@
...
+
+Looking at the contours of the Rosenbrock function, with its prolonged, narrow valley, it is not difficult to imagine that curvature information might make a difference. And, as you'll see in a second, it really does. Before though, one note on the code. When using L-BFGS, it is necessary to wrap function call and gradient evaluation in a closure (`calc_loss()`, in the below snippet), for them to be callable several times per iteration. You can convince yourself that the closure is, in fact, entered repeatedly, by inspecting this code snippet's chatty output:
```
batpigandme, line 185: > it is necessary to wrap function call and gradient evaluation in a closure

Either give this an article (probably the), or pluralize it (function _calls_ and gradient evaluation). "function call" can't really stand alone.

batpigandme: > You can convince yourself
Convince yourself sort of implies that it's not true (we convince ourselves of many false things!). Maybe try something like _You can see for yourself that the closure is…_?

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
```diff
@@ -0,0 +1,640 @@
...
+    Gradient is:  0.0002865837 -0.0001251698 
+
+    Value is:  4.547474e-12 
+    Gradient is:  -1.907349e-05 9.536743e-06 
+
+    Iteration:  3 
+    Value is:  4.547474e-12 
+    Gradient is:  -1.907349e-05 9.536743e-06 
+
+Even though we ran the algorithm for three iterations, the optimal value really is reached after two. Seeing how well this worked, we try L-BFGS on a slightly more crazy function, named *flower*, for pretty self-evident reasons.
```
batpigandme, line 250:
```suggestion
Even though we ran the algorithm for three iterations, the optimal value really is reached after two. Seeing how well this worked, we try L-BFGS on a slightly more complex function, named *flower*, for pretty self-evident reasons.
```

batpigandme: You could use a bunch of different words here (e.g. complex, complicated, elaborate…), but I'd avoid "crazy" both because it's unclear what that means, and because it reads as "ableist" these days: https://www.npr.org/2019/07/08/739643765/why-people-are-arguing-to-stop-using-the-words-crazy-and-insane

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
```diff
@@ -0,0 +1,640 @@
...
+
+## (Yet) more fun with L-BFGS
+
+Here is the *flower* function.[^2] Mathematically, its minimum is near `(0,0)`, but at `(0,0)` itself it is undefined since the `atan2` used in the function is not defined in that location.
```
batpigandme, line 254:
```suggestion
Here is the *flower* function.[^2] Mathematically, its minimum is near `(0,0)`, though the function itself is technically undefined at  `(0,0)`, since the `atan2` used in the function is not defined in that location.
```

batpigandme: Basically a bunch of words to deal with a reflexive reference in English (and avoid "itself it", which just reads awkwardly)

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
```diff
@@ -0,0 +1,640 @@
...
+    Gradient is:  0 0 
+    X is:  -2.826342e+19 -2.503904e+19 
+
+This has been less of a success. At first, loss decreases nicely, but suddenly, the estimate dramatically overshoots, and keeps bouncing between negative and positive outer space ever after.
+
+Luckily, there is something we can do.
+
+### L-BFGS with line search
+
+Taken in isolation, what a Quasi-Newton method like L-BFGS does is determine the best descent direction. However, as we just saw, a good direction is not enough. With the flower function, wherever we are, the optimal path leads to disaster if we stay on it long enough. We thus need an algorithm that carefully evaluates not only where to go, but also, how far.
```
batpigandme, line 381:
```suggestion
Taken in isolation, what a Quasi-Newton method like L-BFGS does is determine the best descent direction. However, as we just saw, a good direction is not enough. With the flower function, wherever we are, the optimal path leads to disaster if we stay on it long enough. Thus, we need an algorithm that carefully evaluates not only where to go, but also, how far.
```

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
```diff
@@ -0,0 +1,640 @@
...
+
+For this reason, L-BFGS implementations commonly incorporate line search, that is, a set of rules indicating whether a proposed step length is a good one, or should be improved upon.
```
batpigandme, line 383:
```suggestion
For this reason, L-BFGS implementations commonly incorporate _line search_, that is, a set of rules indicating whether a proposed step length is a good one, or should be improved upon.
```

batpigandme: I like giving a term some sort of visual emphasis when you're defining it in a sentence (especially when that's not the only thing going on in the sentence).

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
```diff
@@ -0,0 +1,640 @@
...
+    Gradient is:  -3419.361 -8253.501 
+    X is:  -7.404627e-08 3.066708e-08 
+
+It's still not perfect, but a lot better.
+
+Finally, let's go one step further. Can we use `torch` for constrained optimization?
+
+### Quadratic penalty for constrained optimization
+
+In constrained optimization, we still search for a minimum, but that minimum can't reside just anywhere: Its location has to fulfill some number of additional conditions. In optimization lingo, it has to be *feasible*.
```
batpigandme, line 459: Very nicely described.

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
```diff
@@ -0,0 +1,640 @@
...
+
+To illustrate, we stay with the flower function, but add on a constraint: $\mathbf{x}$ has to lie outside a circle of radius $sqrt(2)$, centered at the origin. Formally, this yields the inequality constraint
+
+$$
+2 - {x_1}^2 - {x_2}^2 <= 0
+$$
+
+A way to minimize *flower* and yet, at the same time, honor the constraint is to use a penalty function. With penalty methods, the value to be minimized is a sum of two things: the target function's output and a penalty reflecting potential constraint violation. Use of a *quadratic* *penalty*, for example, results in adding a multiple of the square of the constraint function's output:
```
batpigandme, line 467: Why is flower italicized in the first sentence here?

batpigandme: > A way to minimize *flower* and yet, at the same time, honor the constraint is to use a penalty function

Maybe simplify this sentence a bit?
“One way to minimize flower while still honoring the constraint is to use a penalty function."

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
```diff
@@ -0,0 +1,640 @@
...
+    cat("Penalty: ", current_penalty, "\n")
+    if (current_penalty == 0) break
+    
+    rho <- rho * gamma
+  }
+
+}
+```
+
+`Minimize()`, called from `penalty_method()`, follows the usual proceedings, but now it minimizes the sum of the target and up-weighted penalty function outputs:
```
batpigandme, line 502:
```suggestion
`minimize()`, called from `penalty_method()`, follows the usual proceedings, but now it minimizes the sum of the target and up-weighted penalty function outputs:
```

batpigandme: When referencing code, always keep the casing! (Since R is case sensitive, we, as writers, have to be so, too!)

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
```diff
@@ -0,0 +1,640 @@
...
+
+    Starting step:  22 , rho =  2097152 
+    Iteration:  1 
+    Value:  0.4142135 
+    X:  0.5411962 1.306563 
+    Penalty:  0 
+
+## Conclusion
+
+Summing up, we've gotten a first impression of the effectiveness of `torch`'s L-BFGS optimizer, especially when used with Strong-Wolfe line search. In fact, in numerical optimization -- as opposed to deep learning, where computational speed is much more of an issue -- there probably is hardly ever a reason to *not* use L-BFGS with line search.
```
batpigandme, line 574:
```suggestion
Summing up, we've gotten a first impression of the effectiveness of `torch`'s L-BFGS optimizer, especially when used with Strong-Wolfe line search. In fact, in numerical optimization -- as opposed to deep learning, where computational speed is much more of an issue -- there is hardly ever a reason to *not* use L-BFGS with line search.
```

batpigandme: You don't need probably when you've got "hardly ever," since it's akin to equivocating/qualifying twice.

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
```diff
@@ -0,0 +1,640 @@
...
+
+We've then caught a glimpse of how to do constrained optimization, a task that arises in many real-world applications. In that regard, this post feels a lot more like a beginning than a stock-taking. There is a lot to explore, from general method fit -- when is L-BFGS well suited to a problem? -- via computational efficacy to applicability to different species of neural networks. Needless to say, if this inspires you to run your own experiments, and/or if you use L-BFGS in your own projects, we'd love to hear your feedback!
```
batpigandme, line 576: Lovely little closer 🙂

batpigandme (approved):
I learned a whole heck of a lot while reading this one!

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
```diff
@@ -0,0 +1,640 @@
...
+    self_contained: false
+    toc: true
+preview: images/preview.jpg
+---
+
+```{r setup, include=FALSE}
+knitr::opts_chunk$set(echo = TRUE, eval = FALSE)
+```
+
+So far, all `torch` use cases we've seen have been in deep learning. However, its automatic differentiation feature is useful in other areas, as well. One prominent example is numerical optimization: We can use `torch` to find the minimum of a function.
```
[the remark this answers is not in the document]
skeydan: great! (killing off the "as well") 
I think what you do feels could be likened to sharpening a cross country skiing pole ... ehem, sharpening a knife :-)

and totally makes sense about the "have seen"...

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
skeydan: Yes, I see what you mean... I think I'm doing it because I'm not too sure about where to place "optimizer", ontology-wise ... If I used it without any formatting, it would be as though it existed in natural language, which I think it does not, or not really. So it's a technical term that signifies a very specific thing, in deep learning, while in other areas (like databases), it refers to something totally different ... 

So by formatting it like code I'm kinda referring to a virtual superclass - that does in fact exist, but is not named that way - subsuming objects created by `optim_adam()` etc. ..

I don't really like it, but I don't know of a better way (yet)...

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
skeydan: yeah, I thought it would be fine without because I'm giving its definition as well as a picture, but you're right, a link can never hurt

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
skeydan: thanks!

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
skeydan: italicized because you don't normally normalize flowers :-)))
(meaning, otherwise I would have felt I need the full "flower function" yet again :-)

About the simplification, not sure... I don't really find it complicated and of course, like its more "oral" sound

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
skeydan: oh yeah right ... needs some discipline, starting a sentence lower-case, though :-)

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
skeydan: caught me again (you know what I'm referring to :-))

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
skeydan: :-)

dfalbel:
Looks great to me @skeydan !

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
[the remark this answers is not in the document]
skeydan: Yes, exactly! (One thing I wasn't hinting at was imagination; the problem with deep composition is more like that derivatives get super tedious to compute, so you wouldn't want do them end-to-end... instead what automatic differentiation libraries like torch do is building them up piece-by-piece, and just storing the intermediate updates to the end-to-end derivatives...)

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
skeydan: > very well known initialism

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
skeydan: thank you!

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
skeydan: yeah... the intention was to hint at Adam not being such a good solution here... I don't like it too much either, gonna think ...

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
skeydan: yes, I get it... i guess "more difficult" (without the "slightly") would do as well...

_posts/2021-04-22-torch-for-optimization/torch-for-optimization.Rmd
skeydan: > Either give this an article (probably the), or pluralize it (function calls and gradient evaluation).

I always dislike the many "the"s I have in my English writing (they feel unidiomatic to me)... but I just had an idea: "both" should do it...

skeydan: > Convince yourself sort of implies that it's not true (we convince ourselves of many false things!). 

Honestly, I wouldn't go that far... I mean, you're right, but to me it's pretty synonymous to your alternative ...

skeydan:
thanks so much @batpigandme and @dfalbel for your reviews!!

PR #162, opened 2021-06-16
bump pandoc version to 2.14

yitao-li:
And then massive cache invalidations. Yay.


skeydan (approved):
great, thank you!

PR #176, opened 2021-11-29
[WIP] fix issue with RSS feed

yitao-li:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -61,7 +61,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: /tmp/_posts
-          key: ${{ hashFiles('.github/pandoc-version') }}-${{ hashFiles('.github/R-version') }}-${{ hashFiles('.github/pkg-versions.Rds') }}-${{ hashFiles('.github/pkg-deps.Rds') }}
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -61,7 +61,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: /tmp/_posts
-          key: ${{ hashFiles('.github/pandoc-version') }}-${{ hashFiles('.github/R-version') }}-${{ hashFiles('.github/pkg-versions.Rds') }}-${{ hashFiles('.github/pkg-deps.Rds') }}
+          key: RStudio-AI-Blog-${{ hashFiles('.github/pandoc-version') }}-${{ hashFiles('.github/R-version') }}-${{ hashFiles('.github/pkg-versions.Rds') }}-${{ hashFiles('.github/pkg-deps.Rds') }}
```
yitao-li, line 64: Per discussion, we can try modifying the cache key for build artifacts (which is different from cache key for R packages BTW). This will cause all blog posts to be built from scratch and we can then see whether that fixes the RSS issue.

skeydan (approved):
thanks!!

_site.yml
```diff
@@ -28,7 +28,7 @@ collections:
     subscribe: _subscribe.html
 output: distill::distill_article
 rss:
-  full_content: TRUE
+  full_content: true
```
yitao-li, line 31: https://rstudio.github.io/distill/blog.html says `true` instead of `TRUE` :slightly_smiling_face:

yitao-li:
This change is not fix the underlying problem (yet).
Per discussion on Slack thread, we should probably revisit this later when the RSS issue is fully fixed in the R packages we use for rendering the site.

PR #207, opened 2023-03-22
add post

skeydan:


_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
```diff
@@ -0,0 +1,376 @@
...
+     self$final_linear <- torch::nn_linear(hidden_channels, out_channels)
+   },
+   forward = function(x) {
+     x <- x |>
+       self$conv1() |>
+       (\(.) torch::nnf_layer_norm(., .$shape[2:4]))() |>
+       torch::nnf_relu()
+     for (i in 1:(length(self$convs))) {
+       x <- x |>
+         (\(.) self$convs[[i]](.))() |>
```
dfalbel, line 294: I don't think you need to lambda here
```suggestion
         self$convs[[i]]() |>
```

dfalbel (commented):
Looks great @skeydan ! Very nice :)

skeydan:
Thanks a lot @dfalbel! Happy to hear you like it!

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
```diff
@@ -0,0 +1,374 @@
...
+    self_contained: false
+    toc: true
+preview: images/preview.jpg
+---
+
+```{r setup, include=FALSE}
+knitr::opts_chunk$set(echo = TRUE, eval = FALSE, fig.width = 6, fig.height = 6)
+```
+
+Convolutional neural networks (CNNs) are great -- they're able to detect features in an image no matter where. Well, not exactly. They're not indifferent to just any kind of movement. Shifting up or down, or left or right, is fine; rotating around an axis is not. That's because of how convolution works: traverse by row, then traverse by column (or the other way round). If we want "more" (e.g., successful detection of an upside-down object), we need to extend convolution to an operation that is *rotation-equivariant*. An operation that is *equivariant* to some type of action will not just register the moved feature per se, but also, keep track of which concrete action made it appear where it is.
```
batpigandme, line 27: > they're able to detect features in an image no matter where. Well, not exactly. They're not indifferent to just any kind of movement.

Do you mean where in the image the features are? I think this is what you mean, but I got a little thrown by your use of the word "movement" in the following sentence, since it made me think we were looking at something non-static. 

You might not need to change anything, just wanted to make sure I was following correctly.

batpigandme, line 27:
```suggestion
Convolutional neural networks (CNNs) are great -- they're able to detect features in an image no matter where. Well, not exactly. They're not indifferent to just any kind of movement. Shifting up or down, or left or right, is fine; rotating around an axis is not. That's because of how convolution works: traverse by row, then traverse by column (or the other way round). If we want "more" (e.g., successful detection of an upside-down object), we need to extend convolution to an operation that is *rotation-equivariant*. An operation that is *equivariant* to some type of action will not only register the moved feature per se, but also, keep track of which concrete action made it appear where it is.
```

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
```diff
@@ -0,0 +1,374 @@
...
+
+**This is the second post in a series that introduces group-equivariant CNNs (GCNNs)***.* The [first](https://blogs.rstudio.com/ai/posts/2023-03-09-group-equivariant-cnn-1/) was a high-level introduction to why we'd want them, and how they work. There, we introduced the key player, the symmetry group that specifies which kinds of transformations are to be treated equivariantly. If you haven't, please take a look at that post first, since here I'll make use of terminology and concepts it introduced.
```
batpigandme, line 29:
```suggestion
**This is the second post in a series that introduces group-equivariant CNNs (GCNNs)***.* The [first](https://blogs.rstudio.com/ai/posts/2023-03-09-group-equivariant-cnn-1/) was a high-level introduction to why we'd want them, and how they work. There, we introduced the key player, the symmetry group, which specifies what kinds of transformations are to be treated equivariantly. If you haven't, please take a look at that post first, since here I'll make use of terminology and concepts it introduced.
```

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
```diff
@@ -0,0 +1,374 @@
...
+
+Today, we code a simple GCNN from scratch. Code and presentation tightly follow a [notebook](https://github.com/phlippe/uvadlc_notebooks/blob/master/docs/tutorial_notebooks/DL2/Geometric_deep_learning/tutorial1_regular_group_convolutions.ipynb) provided as part of University of Amsterdam's 2022 [Deep Learning Course](https://uvadl2c.github.io/). They can't be thanked enough for making available such excellent learning materials.
+
+In what follows, my intent is to explain the general thinking, and how the resulting architecture is built up from smaller modules, assigned a clear purpose each. For that reason, I won't reproduce all the code here; instead, I'll make use of the package [`gcnn`](https://github.com/skeydan/gcnn)`.` Its methods are heavily annotated; so to see some details, don't hesitate to look at the code.
```
batpigandme, line 33: > the resulting architecture is built up from smaller modules, assigned a clear purpose each.

This last part ("assigned a clear purpose each") is a bit  awkward. Maybe "each of which is assigned a clear purpose"?

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
```diff
@@ -0,0 +1,374 @@
...
+
+As of today, `gcnn` implements one symmetry group: $C_4$, the one that has been serving as a running example throughout post one. It is straightforwardly extensible, though, making use of class hierarchies throughout.
```
batpigandme, line 35:
```suggestion
As of today, `gcnn` implements one symmetry group: $C_4$, the one that serves as a running example throughout post one. It is straightforwardly extensible, though, making use of class hierarchies throughout.
```

batpigandme: I might mention that C4 stands for cyclic group of the order 4 (or whatever it stands for) when you introduce it, unless C4 is a common abbreviation. I'm sure it's clear from the previous post, but I think it's always good to assume abbreviations aren't in the reader's "working memory."

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
```diff
@@ -0,0 +1,374 @@
...
+```
+
+    torch_tensor
+     0.0000
+     1.5708
+     3.1416
+     4.7124
+    [ CPUFloatType{4} ]
+
+Elements are straightforwardly represented by their respective rotation angles: $0$, $\frac{\pi}{2}$, $\pi$, and $\frac{3 \pi}{2}$.
```
batpigandme, line 60:
```suggestion
Elements are represented by their respective rotation angles: $0$, $\frac{\pi}{2}$, $\pi$, and $\frac{3 \pi}{2}$.
```

batpigandme: "Straightforward" (or it's mouthful of an adverb-form, straightforwardly) is one of those words like clearly, simply, obviously, just, etc. that I think can be cut much of the time. (If it's straightforward, the reader will garner that…otherwise, they feel bad for finding it non-obvious).

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
```diff
@@ -0,0 +1,374 @@
...
+
+# in C_4$left_action_on_H(), H stands for the symmetry group
+C_4$left_action_on_H(torch_tensor(g1)$unsqueeze(1), torch_tensor(g2)$unsqueeze(1))
+```
+
+    torch_tensor
+     4.7124
+    [ CPUFloatType{1,1} ]
+
+What's with the `unsqueeze()`s? Since $C_4$'s ultimate *raison d'être* is to be part of a neural network, `left_action_on_H()` works with batches of elements, not scalar tensors.
```
batpigandme, line 92: I think I'm missing the conclusion here. Does `unsqueeze()` turn batches of elements back into scalar tensors?

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
```diff
@@ -0,0 +1,374 @@
...
+
+To give a real-life -- sort of -- example, say the operation is a measurement. We'd like to measure the height of the person in the black running outfit and green shoes, standing at the foot of the trail, ready to run up the climb. One option we have is to take the measurement, then let the person run up. Our measurement will be as valid up the mountain as it was down here. Alternatively, we might be polite and not make them wait. Once they're up there, we ask them to come down, and when they're back, we measure their height. The result is the same: Body height is equivariant (more than that: invariant, even) to the action of running up or down. (Of course, height is a pretty dull measure. But something more interesting, such as heart rate, would not have worked so well in this example.)
+
+Returning to things implementation, it turns out that group actions are encoded as matrices. There is one matrix for each group element. For $C_4$, the so-called *regular* representation is a rotation matrix:
```
batpigandme, line 98: Do you mean "things" implementation?

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
```diff
@@ -0,0 +1,374 @@
...
+
+$$
+\begin{bmatrix} \cos(\theta) & -\sin(\theta) \\ \sin(\theta) & \cos(\theta) \end{bmatrix}
+$$
+
+In `gcnn`, the function applying that matrix is `C_4$left_action_on_R2()`. Like its sibling, it is designed to work with batches (of group elements as well as $\mathbb{R}^2$ vectors). Technically, what it does is rotate the grid the image is defined on, and then, re-sample the image. To make this more concrete, that method's code looks about as follows.
```
batpigandme, line 104:
```suggestion
In `gcnn`, the function applying that matrix is `left_action_on_R2()`. Like its sibling, it is designed to work with batches (of group elements as well as $\mathbb{R}^2$ vectors). Technically, what it does is rotate the grid the image is defined on, and then, re-sample the image. To make this more concrete, that method's code looks about as follows.
```

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
```diff
@@ -0,0 +1,374 @@
...
+transformed_img[1,..]$permute(c(2, 3, 1)) |> as.array() |> as.raster() |> plot()
+```
+
+```{r, echo=FALSE, eval=TRUE, fig.alt = "Same goat, rotated up by 90 degrees."}
+knitr::include_graphics("images/z2.jpg")
+```
+
+## Step 2: The lifting convolution
+
+We want to make use of existing, efficient `torch` functionality as much as possible. Concretely, we want to use `nn_conv2d()`. What we need, though, is a convolution kernel that's equivariant not just to translation, but also, to the action of $C_4$. This can be achieved by having one kernel each for every possible rotation.
```
batpigandme, line 151:
```suggestion
We want to make use of existing, efficient `torch` functionality as much as possible. Concretely, we want to use `nn_conv2d()`. What we need, though, is a convolution kernel that's equivariant not just to translation, but also to the action of $C_4$. This can be achieved by having one kernel for each possible rotation.
```

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
```diff
@@ -0,0 +1,374 @@
...
+
+Implement that idea is exactly what `LiftingConvolution` does. The principle is the same as before: First, the grid is rotated, and then, the kernel (weight matrix) is re-sampled to the transformed grid.
```
batpigandme, line 153:
```suggestion
Implementing that idea is exactly what `LiftingConvolution` does. The principle is the same as before: First, the grid is rotated, and then, the kernel (weight matrix) is re-sampled to the transformed grid.
```

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
```diff
@@ -0,0 +1,374 @@
...
+
+Why, though, call this a *lifting convolution*? The usual convolution kernel operates on $\mathbb{R}^2$; our extended version operates on combinations of $\mathbb{R}^2$ and $C_4$. In math speak, it has been *lifted* to the [semi-direct product](https://en.wikipedia.org/wiki/Semidirect_product) $\mathbb{R}^2\rtimes C_4$.
```
batpigandme, line 155:
```suggestion
Why, though, call this a *lifting convolution*? The usual convolution kernel operates on $\mathbb{R}^2$; while our extended version operates on combinations of $\mathbb{R}^2$ and $C_4$. In math speak, it has been *lifted* to the [semi-direct product](https://en.wikipedia.org/wiki/Semidirect_product) $\mathbb{R}^2\rtimes C_4$.
```

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
```diff
@@ -0,0 +1,374 @@
...
+  )
+
+x <- torch::torch_randn(c(2, 3, 32, 32))
+y <- lifting_conv(x)
+y$shape
+```
+
+    [1]  2  8  4 28 28
+
+Since, internally, `LiftingConvolution` uses an additional dimension to realize the product of translations and rotations, the output is not four-, but five-dimensional.
```
batpigandme, line 172:
```suggestion
Since, internally, `LiftingConvolution()` uses an additional dimension to realize the product of translations and rotations, the output is not four-, but five-dimensional.
```

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
```diff
@@ -0,0 +1,374 @@
...
+    hidden_channels = 16 # number of channels per group conv layer
+)
+
+img <- torch::torch_randn(c(4, 1, 32, 32))
+cnn(img)$shape
+```
+
+    [1] 4 1
+
+At casual glance, this `GroupEquivariantCNN` looks like any old CNN ... weren't it for the `group` argument.
```
batpigandme, line 214: Again, here, if you're referring to the function, I'd add trailing parentheses.

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
```diff
@@ -0,0 +1,374 @@
...
+
+When now we inspect its output, we see that the additional dimension is gone. That's because after a sequence of group-to-group convolution layers, the module projects down to a representation that, for each batch item, retains channels only. It thus averages not just over locations -- as we normally do -- but over the group dimension as well. A final linear layer will then provide the requested classifier output (of dimension `out_channels`).
```
batpigandme, line 216:
```suggestion
Now, when we inspect its output, we see that the additional dimension is gone. That's because after a sequence of group-to-group convolution layers, the module projects down to a representation that, for each batch item, retains channels only. It thus averages not just over locations -- as we normally do -- but over the group dimension as well. A final linear layer will then provide the requested classifier output (of dimension `out_channels`).
```

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
```diff
@@ -0,0 +1,374 @@
...
+    ) %>%
+    luz::set_opt_hparams(lr = 1e-2, weight_decay = 1e-4) |>
+    luz::fit(train_dl, epochs = 10, valid_data = test_dl) 
+
+```
+
+    Train metrics: Loss: 0.0498 - Acc: 0.9843
+    Valid metrics: Loss: 3.2445 - Acc: 0.4479
+
+Unsurprisingly, accuracy on the test set is not that great.
```
batpigandme, line 329: Unsurprisingly why? Unsurprisingly…without group equivariance? (Ignore if this is obvious to someone who's a more informed reader, which isn't a terribly high bar 😉 )

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
```diff
@@ -0,0 +1,374 @@
...
+```
+
+    Train metrics: Loss: 0.1102 - Acc: 0.9667
+    Valid metrics: Loss: 0.4969 - Acc: 0.8549
+
+For the group-equivariant CNN, accuracies on test and training sets are a lot closer. That is a nice result! Let's wrap up today's exploit resuming a thought from the first, more high-level post.
+
+## A challenge
+
+Going back to the augmented test set, or rather, the samples of digits displayed, we notice a problem. In row two, column four, there is a digit that "under normal circumstances", should be a 9, but, most probably, is an upside-down 6. (To a human, what suggests this is the squiggle-like thing that seems to be found more often with sixes than with nines.) However, you could ask ... does this *have* to be a problem? Maybe the network just needs to learn the subtleties, the kinds of things a human would spot?
```
batpigandme, line 362:
```suggestion
Going back to the augmented test set, or rather, the samples of digits displayed, we notice a problem. In row two, column four, there is a digit that "under normal circumstances", should be a 9, but, most probably, is an upside-down 6. (To a human, what suggests this is the squiggle-like thing that seems to be found more often with sixes than with nines.) However, you could ask: does this *have* to be a problem? Maybe the network just needs to learn the subtleties, the kinds of things a human would spot?
```

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
```diff
@@ -0,0 +1,374 @@
...
+
+The way I view it, all depends on the context: What really should be accomplished, and how an application is going to be used. With digits on a letter, I'd see no reason why a single digit should appear upside-down; accordingly, full rotation equivariance would be counter-productive. In a nutshell, we arrive at the same canonical imperative advocates of fair, just machine learning keep reminding us of:
```
batpigandme, line 364:
```suggestion
The way I view it, it all depends on the context: What really should be accomplished, and how an application is going to be used. With digits on a letter, I'd see no reason why a single digit should appear upside-down; accordingly, full rotation equivariance would be counter-productive. In a nutshell, we arrive at the same canonical imperative advocates of fair, just machine learning keep reminding us of:
```

batpigandme (approved):
Good stuff. Just a few edits and questions to make sure that I'd followed correctly.

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
skeydan: oh yeah, I didn't like that myself :-) Your way is much better!

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
skeydan: Hehe :-) Caught me again :-)
Didn't want to write "simply", but wanted to contrast this with the method that's discussed next (which is less "straightforward") ... Still, you're right, it's totally enough if I say "less straightforward" when I get there!

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
skeydan: exactly!!
You got it though you don't even know torch (or do you?)  :-)))

It's really just for the reader who attentively looks at the code, or runs it - it needed to be there I think, and since the point gets taken up again, I guess it's important enough to not go into a footnote

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
skeydan: @batpigandme can you explain?
I guess I do, but it's not clear? And quotes ...idk? Should I write sth different (less colloquial) entirely?

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
skeydan: Well, this is sort of a corner case, you know ... it's a function, but it creates a module, a neural-network layer ... which we tend to think about as "agents" ... E.g. in Python, it's a class.
So yeah, I considered both ways, but all in all, I think this one should read better ...

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
skeydan: same thing here ... or "about same", actually ... (_about_, since I' talking about the argument even!)
So ... well if I want the first, I need this one for consistency ...

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
skeydan: Yeah ... you're right. But well, thinking of the typical audience ... I agree, I could leave it in I guess

skeydan:
thank you so much @batpigandme , this was great as always!
(((and I'm happy that all in all, these edits are not that un-subtle [[[mandatory indirection here]]])))

_posts/2023-03-27-group-equivariant-cnn-2/group-equivariant-cnn-2.Rmd
skeydan: no ... more like "after this excursion, get back to the code" ...
then probably "things code" isn't better ...

"this implementation" is a bit different in intent, but correct ... since the following sentence is exactly about  _this_ implementation ...  but maybe I'll write "the implementation", that one's totally neutral

PR #218, opened 2023-06-14
Safetensors edits sk2

skeydan:
Looks great! Just a a few edits. (Removed the "safetensors" category, - I cleaned up the variety of categories at some point in the past and would like to keep the current restricted set if possible.)


dfalbel:
Thank you very much @skeydan !

